PR #22, opened 2023-10-11
solana-ibc: add Anchor tests to CI

mina86:


dhruvja:
Any idea on how long it takes?

mina86:
The test is couple minutes with the full Anchor Build taking below ten. Though for some reason GitHub refuses to start the jobs…

mina86:
OK, so timing wise, when those three steps are run one after the
other:

- `anchor build` takes 5′11″,
- `anchor build -- --features mocks` takes 1′22″ and
- `anchor test --skip build` takes 2′56″.

Altogether nine and a half minutes.  That’s IMO long for CI but not
outrageously so.  One way to improve this is by dropping the first
`anchor build` step.  We’re building the codebase with and without
mocks feature in other steps so that should be mostly covered.

Interestingly, using Docker container is actually slower.  Setting up
the container takes about a minute while with out caching setup,
setting up the environment takes just several seconds.  Granted, when
Rust version changes the environment needs to be rebuilt (taking about
15 minutes) but that won’t happen too often.

dhruvja:
So since we are already running `build`, the additional test is 3 minutes which is quite good compared to outrageously long process of above 20 minutes ( `anchor test` ) it used to take before.

PR #64, opened 2023-10-30
Store packets

dhruvja:
Store the packets in a vector so that it can be queried by the relayer. Right now we can store in the private storage and maybe later on, have a seperate PDA for the packets.


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -122,6 +148,7 @@ pub type InnerClient = String; // Serialized
 pub type InnerConnectionEnd = String; // Serialized
 pub type InnerChannelEnd = String; // Serialized
 pub type InnerConsensusState = String; // Serialized
+pub type InnerPacketMsg = Vec<u8>;
```
blasrodri, line 151: why stored it serialized?

dhruvja: Because the `PacketMsg` doesnt implement `AnchorSerialize` and `AnchorDeserialize` macros

dhruvja:
Storing the packets in a seperate PDA so that we dont run out of storage for our private storage.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -63,7 +67,29 @@ pub mod solana_ibc {
...
+                        {
+                            Some(_) => (),
+                            None => {
+                                inner_store.packets.0.push(serialized_packet)
+                            }
+                        }
+                    }
+                    _ => (),
+                }
+                result
```
mina86, line 92: Could we instead generate a new PDA on demand? Something with seed being `["packet", client_id, connection_id, port_id, channel_id, sequence]`.

Also, how are we deleting packets from the list?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
mina86: [AnchorSerialize and AnchorDeserialize are just aliases to Borsh](https://github.com/coral-xyz/anchor/blob/7e5bdce710f9580e1489089d04afb13f71a5d1c7/lang/src/lib.rs#L58) so PacketMsg does actually implement it.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: well we are not deleting the packets from the list. But i think we can delete them once they are acknowledged?

dhruvja: For PDA in demand, i think we can

blasrodri (approved):

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -51,14 +55,33 @@ pub mod solana_ibc {
...
 
         let errors = messages
             .into_iter()
-            .map(|msg| ibc::core::dispatch(&mut store, &mut router, msg))
+            .map(|msg| {
+                let result =
+                    ibc::core::dispatch(&mut store, &mut router, msg.clone());
+                if let MsgEnvelope::Packet(packet) = msg {
+                    // store the packet if not exists
+                    // TODO(dhruvja) Store in a PDA with channelId, portId and Sequence
```
mina86, line 71: Long term we just need to get rid of storing packets in Solana.

mina86 (approved):

PR #66, opened 2023-10-31
solana-ibc: don’t wrap deliver messages in protobuf message

mina86:
Instead of creating an any wrapper for possible deliver messages, pass
the MsgEnvelope directly and encode it with borsh.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -37,7 +37,7 @@ pub mod solana_ibc {
 
     pub fn deliver(
         ctx: Context<Deliver>,
-        messages: Vec<AnyCheck>,
+        messages: Vec<ibc::core::MsgEnvelope>,
     ) -> Result<()> {
         msg!("Called deliver method");
         let _sender = ctx.accounts.sender.to_account_info();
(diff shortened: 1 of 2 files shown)


dhruvja (approved):
LGTM

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -56,19 +56,11 @@ pub mod solana_ibc {
...
-                errors.push(err);
-            }
-        }
+        let errors = messages
+            .into_iter()
+            .map(|msg| ibc::core::dispatch(&mut store, &mut router, msg))
+            .filter_map(core::result::Result::err)
+            .collect::<Vec<_>>();
 
         // Drop refcount on store so we can unwrap the Rc object below.
```
blasrodri, line 65: this comment line is old right? we aren't unwrapping anything

dhruvja: We are unwrapping the store so that we can log it. But we are only doing it for testing so if we dont log it, we can remove the drop.

mina86: `Rc::try_unwrap(store.0)` is what the comment referrs to.

PR #67, opened 2023-10-31
solana-ibc: accept just one message in deliver

mina86:
Change deliver function to take a single message rather than e vector
of messages.  If caller needs to handle multiple messages, which is
probably uncommon, they can batch Solana transactions.

With that change, handle errors better by actually returning them on
failure.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -37,12 +37,10 @@ pub mod solana_ibc {
 
     pub fn deliver(
         ctx: Context<Deliver>,
-        messages: Vec<ibc::core::MsgEnvelope>,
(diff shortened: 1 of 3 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -22,6 +22,7 @@ declare_id!("EnfDJsAK7BGgetnmKzBx86CsgC5kfSPcsktFCQ4YLC81");
 
 mod client_state;
 mod consensus_state;
+mod errors;
```
dhruvja, line 25: did u miss the errors file?

mina86: Kind of.

dhruvja (approved):

PR #72, opened 2023-11-01
Transfer implementation

dhruvja:
Adding methods for transferring the tokens


solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
```diff
@@ -422,7 +422,8 @@ impl ExecutionContext for IbcStorage<'_, '_> {
         seq: Sequence,
     ) -> Result {
         msg!("store_next_sequence_send: path: {path}, seq: {seq}");
-        let store: &mut IbcStorageInner<'_, '_> = &mut self.0.borrow_mut();
+        let store: &mut IbcStorageInner<'_, '_, '_, '_> =
+            &mut self.0.borrow_mut();
```
mina86, line 426:
```suggestion
        let store = self.0.borrow_mut();
```
here and throughout.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -229,16 +234,18 @@ pub struct PrivateStorage {
...
 /// All the structs from IBC are stored as String since they dont implement AnchorSerialize and AnchorDeserialize
 #[derive(Debug)]
-pub struct IbcStorageInner<'a, 'b> {
+pub struct IbcStorageInner<'a, 'b, 'c, 'd> {
     pub private: &'a mut PrivateStorage,
     pub provable:
         solana_trie::AccountTrie<core::cell::RefMut<'a, &'b mut [u8]>>,
+    pub accounts: Vec<AccountInfo<'c>>,
+    pub signer: AccountInfo<'d>,
 }
```
mina86, line 243: I’m fairly certain you can use single lifetime for `'c` and `'d`.  In fact you may be able to use `'b` for those lifetimes.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,81 @@
...
+        msg!(
+            "Sending coins from account {} to account {}, trace path {}, base \
+             denom {}",
+            from,
+            to,
+            amt.denom.trace_path,
+            amt.denom.base_denom
+        );
+        let sender_id = from.to_string();
+        let receiver_id = to.to_string();
```
mina86, line 32: I’m quite suspicious of those and other `to_string` calls in this function. For example comparison in `find` later should just work with no string conversion.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,81 @@
...
 };
 use ibc::applications::transfer::error::TokenTransferError;
 use ibc::applications::transfer::PrefixedCoin;
 use ibc::core::ics24_host::identifier::{ChannelId, PortId};
 use ibc::Signer;
+use uint::FromDecStrErr;
 
 // use crate::module_holder::IbcStorage<'_,'_>;
-use crate::IbcStorage;
+use crate::{id, IbcStorage, MINT_ESCROW_SEED};
```
mina86, line 14: IMO `crate::ID` is better than `crate::id()`. Also, it’s so short I’d consider not `use`ing it and just using full path each place it’s used.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,81 @@
...
+            from,
+            to,
+            amt.denom.trace_path,
+            amt.denom.base_denom
+        );
+        let sender_id = from.to_string();
+        let receiver_id = to.to_string();
+        let base_denom = amt.denom.base_denom.to_string();
+        let amount = amt.amount;
+        // Since amount is u256 which is array of u64, so if the amount is above u64 max, it means that the amount value at index 0 is max.
```
mina86, line 35: This doesn’t track. `[1, 1, 1, 1]` is above `u64::MAX` and none of the elements are `u64::MAX`. I don’t think I understand what this condition is doing.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,81 @@
...
+            ));
+        }
+        let (_token_mint_key, bump) =
+            Pubkey::find_program_address(&[base_denom.as_ref()], &id());
+        let store = self.0.borrow();
+        let sender = store
+            .accounts
+            .iter()
+            .find(|account| account.key.to_string() == sender_id)
+            .ok_or(TokenTransferError::ParseAccountFailure)?;
```
mina86, line 48: This needs to be extracted to a function. Possibly method on IbcStorage

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: We couldn't use same lifetimes since remaining accounts and signer from account struct have different lifetimes. But we are not using signer anyways, so I will remove it. I tried using 'b but it didn't work

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: Maybe I misunderstood how u256 is coded. Suppose I have a number 224, how would it be coded as u256?

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: Sure got it

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
mina86: [U256 uses little-endian](https://docs.rs/primitive-types/0.12.2/primitive_types/struct.U256.html) so 224 would be encoded as `[224, 0, 0, 0]`.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: We have to convert `AccountId` from IBC to string since its just a [wrapper over the string.](https://github.com/cosmos/ibc-rs/blob/main/crates/ibc/src/applications/transfer/context.rs#L25). But for comparing the Pubkeys, we can just do it directly instead of converting.

dhruvja: or better yet, we convert the `AccountId` to string and then to Pubkey. So all comparisions happen with Pubkey and the conversion happens only once.
```suggestion
        let sender_id = Pubkey::from_str(&from.to_string()).unwrap();
        let receiver_id = Pubkey::from_str(&to.to_string()).unwrap();
```

mina86: So that would mean we don’t need to do conversion when we iterate, right? Could we do `as_str()` instead of `to_string()`?

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
mina86: You can do `u64::try_from(amount)`. However, rather than converting to u64 we probably just need to deal with U256 numbers.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: Yes we dont do conversion when we iterate. I dont think we can use `as_str` as the `ibc::Signer` doesnt implement it.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
mina86: `u64::try_from(U256::from(amount))` then. I still don’t see why we need to convert though. Just use Amount or U256 everywhere. Tokens have tendency to use large denominations so u64 may very well be insufficient.

dhruvja: But we need to convert anyhow since, the transfer method only expects u64

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
mina86: It has `as_ref` though.

It might be worth looking into using `Pubkey` as `Self::AccountId`.  Or some kind of wrapper around `Pubkey` since we need a type which has `TryFrom<ibc::Signer>`.

dhruvja: sure sure

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
mina86: I see. So it’s Solana’s limitation which uses u64 for tokens. Could you add a comment explaining that where you check the amount?

dhruvja: Sure sure I will add that.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,78 @@
...
 use ibc::core::ics24_host::identifier::{ChannelId, PortId};
-use ibc::Signer;
+use uint::FromDecStrErr;
+use primitive_types::U256;
 
 // use crate::module_holder::IbcStorage<'_,'_>;
-use crate::IbcStorage;
+use crate::{IbcStorage, MINT_ESCROW_SEED};
+
+pub struct InnerPubkey(pub Pubkey);
```
dhruvja, line 19: This is what u suggested, a wrapped type around `Pubkey` instead of using `ibc::Signer`?

dhruvja: sure sure

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,78 @@
...
+
+impl TryFrom<ibc::Signer> for InnerPubkey {
+
+   type Error = ParsePubkeyError;
```
mina86, line 23: You can also write this as:

```suggestion
   type Error = <Pubkey as FromStr>::Err;
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,78 @@
...
+
+pub struct InnerPubkey(pub Pubkey);
+
+impl TryFrom<ibc::Signer> for InnerPubkey {
+
+   type Error = ParsePubkeyError; 
 
-impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
+   fn try_from(value: ibc::Signer) -> Result<Self, Self::Error> {
+        Ok(InnerPubkey(Pubkey::from_str(&value.to_string())?))
```
mina86:
```suggestion
        Pubkey::from_str(value.as_ref()).map(Self)
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,78 @@
...
+   } 
+}
+
```
mina86, line 29: Some further suggestions regarding the type.  You could also implement
PartialEq:

```
impl PartialEq<Pubkey> for InnerPubkey {
    #[inline]
    fn eq(&self, rhs: &Pubkey) -> bool { &self.0 == rhs }
}

impl PartialEq<InnerPubkey> for Pubkey {
    #[inline]
    fn eq(&self, rhs: &InnerPubkey) -> bool { self == &rhs.0 }
}
```

This way you won’t need to access the wrapped value when comparing.

You can also further add some more derives:

```
#[derive(Clone, PartialEq, Eq, derive_more::From, derive_more::Into)]
#[into(owned, ref, ref_mut)]
pub struct AccountId(Pubkey);
```

Perhaps I’m just being pedantic, but to me `Pubkey::from(account)`
reads better than `account.0`.

If you really want to go to town with this type, you can also do:

```
impl core::fmt::Debug for AccountId {
    #[inline]
    fn fmt(&self, fmtr: &mut core::fmt::Formatter) -> core::fmt::Result {
        self.0.fmt(fmtr)
    }
}

impl core::fmt::Display for AccountId {
    #[inline]
    fn fmt(&self, fmtr: &mut core::fmt::Formatter) -> core::fmt::Result {
        self.0.fmt(fmtr)
    }
}
```

and you’d probably never need `.0`.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,78 @@
...
+pub struct AccountId(Pubkey);
+
+impl TryFrom<ibc::Signer> for AccountId {
+
+   type Error = ParsePubkeyError; 
 
-impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
+   fn try_from(value: ibc::Signer) -> Result<Self, Self::Error> {
+        Ok(AccountId(Pubkey::from_str(&value.to_string())?))
```
mina86, line 26:
```suggestion
        Pubkey::from_str(value.as_ref()).map(Self)
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -108,3 +221,21 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
 }
+
+fn get_account_info_from_key<'a, 'b>(accounts: &'a Vec<AccountInfo<'b>>, key: Pubkey) -> Result<&'a AccountInfo<'b>, TokenTransferError> {
+    accounts
+    .iter()
+    .find(|account| account.key == &key)
+    .ok_or(TokenTransferError::ParseAccountFailure)
+}
+
+fn check_amount_overflow(amount: Amount) -> Result<(), TokenTransferError> {
```
mina86, line 232: This should return `u64` on success so the caller can use this function to convert to `u64`.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -108,3 +221,21 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+    // Solana transfer only supports u64 so checking if the token transfer amount overflows. If it overflows we return an error
+    // Since amount is u256 which is array of u64, so if the amount is above u64 max, it means that the amount value at index 0 is max.
```
mina86, line 234: This should be a docstring for the function.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -108,3 +221,21 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+    if amount[0] == u64::MAX {
+        return Err(TokenTransferError::InvalidAmount(
+            FromDecStrErr::InvalidLength,
+        ));
+    }
+    Ok(())
```
mina86, line 240:
```suggestion
    u64::try_from(U256::from(amount)).map_err(|_| {
        TokenTransferError::InvalidAmount(
            FromDecStrErr::InvalidLength
        )
    })
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -108,3 +221,21 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
         Ok(())
     }
 }
+
+fn get_account_info_from_key<'a, 'b>(accounts: &'a Vec<AccountInfo<'b>>, key: Pubkey) -> Result<&'a AccountInfo<'b>, TokenTransferError> {
```
mina86, line 225: Prefer taking slice arguments rather then reference to vector.

```suggestion
fn get_account_info_from_key<'a, 'b>(accounts: &'a [AccountInfo<'b>], key: Pubkey) -> Result<&'a AccountInfo<'b>, TokenTransferError> {
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: But `Pubkey::from(account)` would require me to implement a trait on Pubkey but since it is from external crate, it wouldnt be possible for me to do it. I can convert from `Pubkey` to `AccountId`, not the other way around i suppose.

dhruvja: But if you wanna remove the `.0`, what do u think about this?
```suggestion
impl AccountId {
    fn to_pubkey(self) -> Pubkey {
        self.0
    }
}
```

mina86: > But Pubkey::from(account) would require me to implement a trait on Pubkey but since it is from external crate,

That’s fine. You can implement `From<InternalType> for ExternalType` traits.  `derive_more::Into` will do that for you.

dhruvja: alright

blasrodri:
where are we at with this?

dhruvja:
> where are we at with this? 

Writing some mocks, so that I can test the transfer. Once that's done we can merge. Sorry for the delay.

blasrodri:
@dhruvja we need to deal with SPL tokens and native SOL for transfer.

dhruvja:
> @dhruvja we need to deal with SPL tokens and native SOL for transfer.

I have the current implementation for SPL tokens, if it works we can implement the same for native SOL.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -160,6 +191,152 @@ pub mod solana_ibc {
...
+        port_id: PortId,
+        _channel_id: ChannelId,
+        _base_denom: String,
+        commitment_prefix: CommitmentPrefix,
+        client_id: ClientId,
+        counterparty_client_id: ClientId,
+    ) -> Result<()> {
+        if !TEST {
+            panic!();
+        }
```
mina86, line 207: Why not guard the whole method with `#[cfg(feature = "mocks")]`?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -227,12 +405,82 @@ pub struct Deliver<'info> {
...
+        let code = ErrorDiscriminants::from(err) as u32;
+        anchor_lang::error::ERROR_CODE_OFFSET + code
+    }
+}
+
+#[event]
+pub struct EmitIBCEvent {
+    pub ibc_event: Vec<u8>,
+}
+
```
mina86, line 482:
```suggestion
```

Those now live in error.rs and events.rs.  This looks like left over from previous version of the code.

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -54,6 +73,66 @@ macro_rules! make_message {
...
+
+        self.remaining_accounts.iter().for_each(|&account| {
+            accounts.push(AccountMeta {
+                pubkey: account,
+                is_signer: false,
+                is_writable: false,
+            })
+        });
+
+        accounts
```
mina86, line 132: Isn’t there some Anchor derive that can do this for us?  Can’t we
`#[derive(Account)]` or something on the struct?

---

As per existing code, a few small things for minor optimisations.

Firstly, you can do `[...].to_vec()` to avoid the repetition of
`accounts.push`.  I think it ends up looking cleaner and it also
potentially works better with minimising allocations (though to be
honest I dunno what’s vector’s first allocation size so this may be
a moot point).

Secondly, I would suggest avoiding `for_each` function.  It’s nearly
always better to use an explicit loop.  `for_each` rarely gives you
anything and it introduces a context where keywords change meaning
(`return` is now `break`, `break` is not valid) making the code harder
to understand.

And since you’re pushing to a vector, it’s probably better to use
`Vec::extend` anyway.

But overall, if you really care about micro optimisations, this whole
thing can be written with iterator chaining and a single collect
I believe:

```suggestion
        let accounts = [
            AccountMeta {
                pubkey: self.sender,
                is_signer: true,
                is_writable: true,
            },
            AccountMeta {
                pubkey: self.storage,
                is_signer: false,
                is_writable: true,
            },
            AccountMeta {
                pubkey: self.trie,
                is_signer: false,
                is_writable: true,
            },
            AccountMeta {
                pubkey: self.packets,
                is_signer: false,
                is_writable: true,
            },
            AccountMeta {
                pubkey: self.chain,
                is_signer: false,
                is_writable: true,
            },
            AccountMeta {
                pubkey: self.system_program,
                is_signer: false,
                is_writable: false,
            },
        ];

        let remaining = self
            .remaining_accounts
            .iter()
            .map(|&account| AccountMeta {
                pubkey: account,
                is_signer: false,
                is_writable: false,
            });

        accounts.into_iter().chain(remaining).collect::<Vec<_>>()
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: understood

solana/solana-ibc/programs/solana-ibc/src/tests.rs
dhruvja: right right, that makes total sense.

dhruvja: > Isn’t there some Anchor derive that can do this for us? Can’t we
#[derive(Account)] or something on the struct?

We have to use our own struct since we need to pass the remaining accounts. The anchor derive doesnt enable us to do that.

mina86:
Did a merge though I’m not quite sure what’s the appropriate way to do time and height updates in mock_deliver.

dhruvja:
> Did a merge though I’m not quite sure what’s the appropriate way to do time and height updates in mock_deliver.

For mock, i dont think it matters much. Because when using tendermint, when we create a client it updates height and time automatically. I wonder why it doesnt do on mock. So i do it on mock deliver. But I will look at the best solution once.

mina86 (commented):
As discussed, consider moving the body of mock_delivery to a separate module so there’s no need for all the noisy cfg-gated use statements. I haven’t yet carefully looked at all the business logic but overall looks good to me.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -29,13 +109,47 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
         amt: &PrefixedCoin,
     ) -> Result<(), TokenTransferError> {
         msg!(
-            "Minting coins for account {}, trace path {}, base denom {}",
+            "Minting coins for account {:?}, trace path {}, base denom {}",
```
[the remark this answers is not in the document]
dhruvja: no i changed it because i was facing some error and wanted to debug. I will remove it.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -29,13 +109,47 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
...
             account,
             amt.denom.trace_path,
             amt.denom.base_denom
         );
+        msg!("This should appear before");
```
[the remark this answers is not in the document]
dhruvja: will remove it.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,209 @@
...
+        if matches!(from, AccountIdx::Escrow(_)) {
+            let (mint_authority_key, mint_authority_bump) =
+                Pubkey::find_program_address(&[MINT_ESCROW_SEED], &crate::ID);
+
+            let mint_authority =
+                get_account_info_from_key(accounts, mint_authority_key)?;
+
+            let bump_vector = mint_authority_bump.to_le_bytes();
+            let inner = vec![MINT_ESCROW_SEED, bump_vector.as_ref()];
+            let outer = vec![inner.as_slice()];
```
mina86, line 159: No need for vectors:

```suggestion
            let seeds = [MINT_ESCROW_SEED, bump_vector.as_ref()];
            let seeds = seeds.as_ref();
            let seeds = core::slice::from_ref(&seeds);
```

Same in the else branch.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,209 @@
...
+                authority: authority.clone(),
+            };
+            let cpi_ctx = CpiContext::new_with_signer(
+                token_program.clone(),
+                transfer_instruction,
+                outer.as_slice(), //signer PDA
+            );
+
+            anchor_spl::token::transfer(cpi_ctx, amount_in_u64).unwrap();
+        }
```
mina86, line 204: There seems to be quite a bit of common code between both branches.  Isn’t this correct:

```suggestion
        let (mint_authority_key, mint_authority_bump) =
            Pubkey::find_program_address(&[MINT_ESCROW_SEED], &crate::ID);

        let seeds = [MINT_ESCROW_SEED, core::slice::from_ref(&mint_authority_bump)];
        let seeds = seeds.as_ref();
        let seeds = core::slice::from_ref(&seeds);

        let authority = if matches!(from, AccountIdx::Escrow(_)) {
            get_account_info_from_key(accounts, mint_authority_key)?
        } else {
            let sender_token_account =
                TokenAccount::try_from_slice(&sender.data.borrow()[..])?;
            get_account_info_from_key(accounts, sender_token_account.owner)?
        }

        let transfer_instruction = Transfer {
            from: sender.clone(),
            to: receiver.clone(),
            authority: authority.clone(),
        };
        let cpi_ctx = CpiContext::new_with_signer(
            token_program.clone(),
            transfer_instruction,
            seeds,
        );
        anchor_spl::token::transfer(cpi_ctx, amount)?;
```

Also, probably should prefer returning errors to `unwrap`.

solana/solana-ibc/programs/solana-ibc/src/mocks.rs
```diff
@@ -0,0 +1,199 @@
...
+        host_head,
+    });
+
+    let any_client_state = store.client_state(&client_id).unwrap();
+    let client_state: MockClientState = match any_client_state {
+        client_state::AnyClientState::Tendermint(_) => {
+            panic!("Only for mocks")
+        }
+        client_state::AnyClientState::Mock(mock) => mock,
+    };
```
mina86, line 65:
```suggestion
    let client_state = MockClientState::try_from(any_client_state).unwrap();
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -7,14 +7,22 @@ extern crate alloc;
...
 
-pub const CHAIN_SEED: &[u8] = b"chain";
-pub const PACKET_SEED: &[u8] = b"packet";
-pub const SOLANA_IBC_STORAGE_SEED: &[u8] = b"private";
-pub const TRIE_SEED: &[u8] = b"trie";
+const CHAIN_SEED: &[u8] = b"chain";
+const PACKET_SEED: &[u8] = b"packet";
+const SOLANA_IBC_STORAGE_SEED: &[u8] = b"private";
+const TRIE_SEED: &[u8] = b"trie";
+const MINT_ESCROW_SEED: &[u8] = b"mint_escrow";
```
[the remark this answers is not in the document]
dhruvja: Yes its needed, my bad. Maybe i missed it when i pulled from master.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,209 @@
...
+use strum::Display;
+use uint::FromDecStrErr;
 
-// use crate::module_holder::IbcStorage<'_,'_>;
+use crate::storage::ids::PortChannelPK;
 use crate::storage::IbcStorage;
+use crate::MINT_ESCROW_SEED;
+
+#[derive(Clone, PartialEq, Eq, derive_more::From)]
+pub struct AccountId(Pubkey);
```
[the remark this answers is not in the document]
dhruvja: No i dont think we need it.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,209 @@
...
+        self.0.fmt(fmtr)
+    }
+}
+
+impl From<&AccountId> for Pubkey {
+    fn from(value: &AccountId) -> Self { value.0 }
+}
+
+/// Structure to identify if the account is escrow or not. If it is escrow account, we derive the escrow account using port-id, channel-id and denom.
+#[derive(Clone, Display, PartialEq, Eq, derive_more::From)]
```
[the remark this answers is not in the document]
dhruvja: I added this but i removed the custom error because i dont think we were using it anyhow.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -149,14 +158,44 @@ pub mod solana_ibc {
...
+        // msg!("This is final structure {:?}", inner.private);
 
         // msg!("this is length {}", TrieKey::ClientState{ client_id: String::from("hello")}.into());
 
         Ok(())
     }
+
+    /// Called to set up a connection, channel and store the next
+    /// sequence.  Will panic if called without `mocks` feature.
+    #[allow(unused_variables)]
```
mina86, line 172:
```suggestion
```
This shouldn’t be necessary.

solana/solana-ibc/programs/solana-ibc/src/storage/ids.rs
```diff
@@ -160,8 +160,10 @@ impl PortChannelPK {
         })
     }
 
+    #[allow(dead_code)]
     pub fn port_id(&self) -> ibc::PortId { ibc::PortId::from(&self.port_key) }
 
+    #[allow(dead_code)]
```
mina86, line 166:
```suggestion
    pub fn port_id(&self) -> ibc::PortId { ibc::PortId::from(&self.port_key) }

```
I don’t think those are necessary either.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,131 @@
...
+    Signer(Pubkey),
+    Escrow(PortChannelPK),
+}
+
+impl TryFrom<ibc::Signer> for AccountId {
+    type Error = <Pubkey as FromStr>::Err;
+
+    fn try_from(value: ibc::Signer) -> Result<Self, Self::Error> {
+        Ok(Pubkey::try_from(value.as_ref()).map(Self::Signer).unwrap())
+    }
```
mina86, line 35: This is confusing.  For one why `unwrap` rather than returning the
error?  Furthermore, you’re using Pubkey::try_from but taking error
from FromStr trait.  Perhaps:

```suggestion
    type Error = <Pubkey as FromStr>::Err;

    fn try_from(value: ibc::Signer) -> Result<Self, Self::Error> {
        Pubkey::from_str(value.as_ref()).map(Self::Signer)
    }
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -1,26 +1,131 @@
...
-impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
+impl TryFrom<&AccountId> for Pubkey {
+    type Error = ();
+
+    fn try_from(value: &AccountId) -> Result<Self, Self::Error> {
+        match value {
+            AccountId::Signer(signer) => Ok(*signer),
+            AccountId::Escrow(_) => Err(()),
+        }
+    }
```
mina86, line 62: I think it makes sense to take advantage of existing TryFrom
implementation that derive_more gives us:

```suggestion
    type Error = <Self as TryFrom<AccountId>>::Error;;

    fn try_from(value: &AccountId) -> Result<Self, Self::Error> {
        Self::try_from(*value)
    }
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -34,8 +139,41 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
             amt.denom.trace_path,
             amt.denom.base_denom
         );
+        let receiver_id = account
+            .try_into()
+            .map_err(|_| TokenTransferError::ParseAccountFailure)?;
+        let base_denom = amt.denom.base_denom.to_string();
```
mina86, line 145:
```suggestion
        let base_denom = amt.denom.base_denom.as_str();
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -34,8 +139,41 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
...
+        let accounts = &store.accounts;
+        let receiver = get_account_info_from_key(accounts, receiver_id)?;
+        let token_mint = get_account_info_from_key(accounts, token_mint_key)?;
+        let token_program = get_account_info_from_key(accounts, spl_token::ID)?;
+        let mint_authority =
+            get_account_info_from_key(accounts, mint_authority_key)?;
+
+        let bump_vector = mint_authority_bump.to_le_bytes();
+        let inner = vec![MINT_ESCROW_SEED, bump_vector.as_ref()];
+        let outer = vec![inner.as_slice()];
```
mina86, line 162: No need for vec as before. We probably need a function for this. One which takes reference to `mint_authority_bump` and returns `&[&[&[u8]]]` for the seeds.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -50,14 +188,46 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
             amt.denom.trace_path,
             amt.denom.base_denom
         );
+        let burner_id = account
+            .try_into()
+            .map_err(|_| TokenTransferError::ParseAccountFailure)?;
+        let base_denom = amt.denom.base_denom.to_string();
```
mina86, line 194:
```suggestion
        let base_denom = amt.denom.base_denom.as_str();
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -50,14 +188,46 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
...
+        let accounts = &store.accounts;
+        let burner = get_account_info_from_key(accounts, burner_id)?;
+        let token_mint = get_account_info_from_key(accounts, token_mint_key)?;
+        let token_program = get_account_info_from_key(accounts, spl_token::ID)?;
+        let mint_authority =
+            get_account_info_from_key(accounts, mint_authority_key)?;
+
+        let bump_vector = bump.to_le_bytes();
+        let inner = vec![base_denom.as_ref(), bump_vector.as_ref()];
+        let outer = vec![inner.as_slice()];
```
mina86, line 210: ditto seeds

mina86 (approved):
Looks good enough for testing now though there is a lot of `unwrap` which should instead propagate the error.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: Yes i think the above makes sense.

PR #73, opened 2023-11-01
solana-ibc: add function for checking the Trie account

mina86:
Extract code which creates AccountTrie from the AccountInfo to
a separate function so that it can be easily reused.  Furthermore, add
ownership checks which so far were missing.

Long term I’d like to have a wrapper which can be used inside of
Accounts struct but I couldn’t quite figure out the proper way to
implement it.  AccountTrie isn’t quite zero copy (since it makes copy
of the header data).  This is left for future improvement.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/storage.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/storage.rs
@@ -129,13 +129,41 @@ pub(crate) struct PrivateStorage {
     pub ibc_events_history: BTreeMap<InnerHeight, Vec<InnerIbcEvent>>,
 }
 
+/// Provable storage, i.e. the trie, held in an account.
+pub type AccountTrie<'a, 'b> =
+    solana_trie::AccountTrie<core::cell::RefMut<'a, &'b mut [u8]>>;
+
+/// Checks contents of given unchecked account and returns a trie if it’s valid.
+///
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -129,13 +129,41 @@ pub(crate) struct PrivateStorage {
...
+    solana_trie::AccountTrie<core::cell::RefMut<'a, &'b mut [u8]>>;
+
+/// Checks contents of given unchecked account and returns a trie if it’s valid.
+///
+/// The account needs to be owned by [`crate::ID`] and
+pub fn get_provable_from<'a, 'info>(
+    info: &'a UncheckedAccount<'info>,
+    name: &str,
+) -> Result<AccountTrie<'a, 'info>> {
+    fn get<'a, 'info>(
```
dhruvja, line 143: Why do we have this code block as a method when we are using it only once?

mina86: It’s going to be used in other places in the future.

Still, even though it’s used once, it’s a good unit of code to extract.

dhruvja: Alright, makes sense.

dhruvja (approved):

PR #74, opened 2023-11-01
solana-ibc: store and read commitments from the trie

mina86:
Firstly, since packet and ack commitments are 32-byte long hashes,
store them directly in the provable trie rather than storing their
hash.

Secondly, implement retrieval of the commitments properly.  Rather
than trying to interpret sequences as commitment, read the hashes from
the provable storage.

Lastly, get rid of sequence_sets maps in the private store which don’t
serve any actual purpose.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
@@ -1,5 +1,3 @@
-use std::collections::BTreeMap;
-
 use anchor_lang::emit;
 use anchor_lang::prelude::borsh;
 use anchor_lang::solana_program::msg;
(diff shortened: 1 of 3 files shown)


dhruvja:
I think we would also need to store the packet commitments and acks. Something like this for the moment
```
pub packet_commitments: BTreeMap<(PortId, ChannelId, Sequence), Vec<u8>>,
pub packet_acks: BTreeMap<(PortId, ChannelId, Sequence), Vec<u8>>,
```

mina86:
They are stored in the trie. We don’t need to store them in private storage.

dhruvja:
> They are stored in the trie. We don’t need to store them in private storage.

I thought u were hashing it. If you are just storing it as it is, i can just fetch it from the trie

dhruvja (approved):

mina86:
`<&CryptoHash>::try_from(commitment.as_ref()).unwrap();` interprets the commitment as 32-byte hash which is then stored directly.

PR #78, opened 2023-11-02
solana-ibc: remember sequence numbers of commitments in private storage

mina86:
For the sake of relayer, remember sequence numbers of packet and
acknowledgement commitments in the private storage.

This isn’t strictly necessary since the data can be read from the
trie, but that requires iterator interface which isn’t yet
implemented.  For the time being, use sets in the private storage.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
@@ -1,3 +1,5 @@
+use alloc::collections::BTreeMap;
+
 use anchor_lang::emit;
 use anchor_lang::prelude::borsh;
 use anchor_lang::solana_program::msg;
@@ -23,7 +25,7 @@ use lib::hash::CryptoHash;
 
 use crate::client_state::AnyClientState;
 use crate::consensus_state::AnyConsensusState;
-use crate::storage::IbcStorage;
+use crate::storage::{self, IbcStorage};
 use crate::trie_key::TrieKey;
 use crate::EmitIBCEvent;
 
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
```diff
@@ -238,17 +240,31 @@ impl ExecutionContext for IbcStorage<'_, '_> {
         commitment: PacketCommitment,
     ) -> Result {
         msg!("store_packet_commitment({path}, {commitment:?})");
+        let mut store = self.borrow_mut();
         let trie_key = TrieKey::from(path);
         // PacketCommitment is always 32-byte long.
         let commitment = <&CryptoHash>::try_from(commitment.as_ref()).unwrap();
-        self.borrow_mut().provable.set(&trie_key, commitment).unwrap();
+        store.provable.set(&trie_key, commitment).unwrap();
+        record_packet_sequence(
```
blasrodri, line 248: This could fail right?

mina86: The record_packet_sequence call? No. It just adds value to a vector.

blasrodri: But don't we have limits on space?

mina86: Yes, in this sense it can fail indeed. We’ve been discussing this quite a bit actually and the current thinking is that this should be fine. Commitments are removed when packets are acknowledged which means that size of this list is equal to number of unacknowledged packets. Similarly, commitments are removed from the trie so its size should be fine as well.

The bigger issue is acknowledgement commitments which are never deleted. However, we believe that we can simply ignore acks and not store them at all.

Lastly, there are also receipts but those are held as sealed in the trie so they should not increase storage size only.

Long story short, for the time being this should be fine but we will definitely come back to it.

blasrodri (approved):

PR #81, opened 2023-11-07
blockchain: add Verifier trait

mina86:
Add an Verifier trait rather than a verify method on a Signature.  The
purpose is to allow additional context to be provided when checking
signatures.  This is needed in Solana where signature checking
requires access to per-request data (specifically instructions native
program which is passed with the rest of accounts).

### DIFF
--- a/common/blockchain/src/manager.rs
+++ b/common/blockchain/src/manager.rs
@@ -199,26 +199,27 @@ impl<PK: crate::PubKey> ChainManager<PK> {
         &mut self,
         pubkey: PK,
         signature: &PK::Signature,
+        verifier: &impl crate::Verifier<PK>,
     ) -> Result<bool, AddSignatureError> {
         let pending = self
             .pending_block
             .as_mut()
             .ok_or(AddSignatureError::NoPendingBlock)?;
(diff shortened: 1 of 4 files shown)


common/blockchain/src/manager.rs
```diff
@@ -199,26 +199,27 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
             .ok_or(AddSignatureError::BadValidator)?
             .stake()
             .get();
-        assert!(pending.signers.insert(pubkey));
+        if !verifier.verify(pending.hash.as_slice(), &pubkey, signature) {
+            return Err(AddSignatureError::BadSignature);
+        }
+
+        if !pending.signers.insert(pubkey) {
+            return Ok(false);
```
blasrodri, line 219: what does `Ok(false)` signals?

mina86: /// Returns `true` if quorum has been reached and the pending block has
    /// graduated to the current block.

`Ok(false)` means that the signature was correct but the block is
still pending because quorum hasn’t been reached.

blasrodri (approved):

PR #84, opened 2023-11-08
Enable stress tests on Miri and shrink couple other tests

mina86:
Limit stress tests to just five iterations when running on Miri.  This
makes them actually quite fast (faster than some regular tests) so
they can be added back to CI.

Furthermore, disable the slowest trie tests on Miri introducing their
smaller variants which run on Miri in reasonable time.  Largest
example is test_del which on my machine takes 144 seconds while
test_del_small takes 18.  To be fair I’d argue it’s still considerable
for a unit test but definitely much more reasonable.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -67,4 +67,4 @@ jobs:
           components: miri
 
       - name: Run tests with Miri
-        run: cargo miri test -- --skip ::stress_test --skip ::anchor
+        run: cargo miri test -- -Z unstable-options --report-time --skip ::anchor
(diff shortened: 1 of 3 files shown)


dhruvja:
Since the building of solana program takes time anyways, do we need to limit the tests?

mina86:
For stress tests if they run the same number of iterations as with non-miri execution, they would take hours to finish so yeah, we need to limit those.  But even for the smaller tests, some CI steps taking long should not be an excuse for a particular step to take long.  Rather the argument should be if we can make SBF building faster.

dhruvja:
> For stress tests if they run the same number of iterations as with non-miri execution, they would take hours to finish so yeah, we need to limit those. But even for the smaller tests, some CI steps taking long should not be an excuse for a particular step to take long. Rather the argument should be if we can make SBF building faster.

Right right, that makes sense.

dhruvja (approved):

PR #86, opened 2023-11-09
solana-ibc: refactor events handling

mina86:
Firstly, remove ibc_events_history from the storage.  The field only
ever grows which isn’t sustainable.  The contract has no use for the
field and off-chain customers should look at history of Solana logs
rather than looking at the field.

Secondly, encode the IBC event directly in an event struct rather than
first borsh serialising it and then emitting that.  Since emitting
causes another borsh serialisation this leads to somewhat awkward and
unnecessary double serialisation.

Lastly, use Solana logging directly rather than using Anchor’s logging
infrastructure.  Anchor really doesn’t give us much as far as I can
tell and having an enum with all possible events which is serialised
is just as if not more convenient.


dhruvja (approved):
LGTM

PR #88, opened 2023-11-13
solana-ibc: add  `build-sbf` to GitHub Actions

dhruvja:
Adding `cargo build-bpf` which uses the `solana-cli` to run the program in solana environment. Enables us to catch any bugs while building the solana program. Ideally, we should run the tests which requires `anchor cli` and takes quite a bit of time to install and run the tests. So for the time being, building the program successfully using `bpf` would suffice.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -50,7 +50,8 @@ jobs:
           components: rustfmt
 
       - name: Install Solana CLI
-        run: set -eux
+        run: | 
+          set -eux
           curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
 
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -54,6 +54,7 @@ jobs:
           set -eux
           curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
+          cp solana/solana-ibc/keypair.json ~/.config/solana/id.json
 
       - name: Run build-bpf
         run: cargo build-sbf
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -52,6 +52,10 @@ jobs:
       - name: Install Solana CLI
         run: curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
         shell: bash
+
+      - name: Adding Solana CLI to the path
+        run: echo "/home/runner/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
+        shell: bash
       
       - name: Run build-bpf
         run: cargo build-bpf
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - master
 
+env:
+  SOLANA_VERSION: v1.16.3
+
 jobs:
   misc:
     name: Miscellaneous checks
@@ -34,6 +37,24 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-features -- -D warnings
+
+  solana-build: 
+    name: Cargo Build using bpf (solana env)
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          components: rustfmt
+
+      - name: Install Solana CLI
+        run: curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
+        shell: bash
+      
+      - name: Run build-bpf
+        run: cargo build-bpf
 
   stable:
     name: Rust tests
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -39,7 +39,7 @@ jobs:
           args: --all-features -- -D warnings
 
   solana-build: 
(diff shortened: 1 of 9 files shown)


dhruvja:
The CI should have failed, maybe its not working as expected.

.github/workflows/master.yml
```diff
@@ -35,6 +38,28 @@ jobs:
...
+          toolchain: stable
+          components: rustfmt
+
+      - name: Install Solana CLI
+        run: curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
+        shell: bash
+
+      - name: Adding Solana CLI to the path
+        run: echo "/home/runner/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
+        shell: bash
```
mina86, line 58: Probably just me but I find the single `echo` instruction as separate
step triggering. ;)  Maybe join them?

```suggestion
      - name: Install Solana CLI
        run: |
          set -eux
          curl -sSfL https://release.solana.com/$SOLANA_VERSION/install | sh
          echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
```

Also note that `bash` is the default shell on Linux so no need to specify it.

.github/workflows/master.yml
```diff
@@ -35,6 +38,28 @@ jobs:
...
+      
+      - name: Run build-bpf
+        run: cargo build-bpf
```
mina86, line 61: Do you need to specify project here?

```suggestion
        run: cargo build-bpf -p solana-ibc
```

.github/workflows/master.yml
dhruvja: alright understood, lemme try that.

.github/workflows/master.yml
dhruvja: `cargo build-bpf` doesnt take `-p` as argument. But running `cargo build-bpf` fails on local so it should fail here too.

mina86: How does it know which package to build? Do we need to `cd solana/solana-ibc` or something?

dhruvja: its like `cargo build`, it picks it up from the cargo workspace.

.github/workflows/master.yml
dhruvja: looks like the above didnt work.

.github/workflows/master.yml
mina86: > But running cargo build-bpf fails on local so it should fail here too.

For me `cargo build-bpf` and `cargo build-bpf -- -p solana-ibc` work fine.

For specifying project, it’s actually `cargo build-bpf -- -p solana-ibc`. We care only about that project so no point trying to build others using bpf.

Lastly, should we use `build-sbf` instead?

dhruvja: Did the build fail for you in local?

> Lastly, should we use build-sbf instead?

Yes, we should since the bpf is deprecated.

mina86: > Did the build fail for you in local?

No, it works just fine:

```
$ cargo build-sbf -- -p solana-ibc
Compiling cfg-if v1.0.0
...
    Finished release [optimized] target(s) in 49.56s
```

.github/workflows/master.yml
```diff
@@ -35,6 +38,28 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-features -- -D warnings
 
+  solana-build: 
+    name: Cargo Build using bpf (solana env)
```
mina86, line 42:
```suggestion
    name: Build SBF
```

.github/workflows/master.yml
dhruvja: Thats strange, because it fails for me. So i had fixed it in another PR which is merged with the master. But dont know why its not failing here or locally for u? Is it smtg wrong on my local?

mina86: Oh, OK, I haven’t tested this PR and yes it does fail:

```
$ cargo build-sbf -- -p solana-ibc
...
error[E0412]: cannot find type `Vec` in this scope
   --> common/lib/src/hash.rs:186:29
    |
186 |     pub(super) struct Inner(Vec<u8>);
    |                             ^^^ not found in this scope
    |
help: consider importing one of these items
    |
186 |     use alloc::vec::Vec;
    |
186 |     use borsh::maybestd::vec::Vec;
    |

error[E0425]: cannot find value `bytes` in this scope
   --> common/lib/src/hash.rs:196:43
    |
196 |             solana_program::hash::hashv(&[bytes]).to_bytes()
    |                                           ^^^^^ not found in this scope

error[E0277]: the trait bound `Inner: Default` is not satisfied
   --> common/lib/src/hash.rs:211:20
    |
210 | #[derive(Default)]
    |          ------- in this derive macro expansion
211 | pub struct Builder(builder::Inner);
    |                    ^^^^^^^^^^^^^^ the trait `Default` is not implemented for `Inner`
    |
    = note: this error originates in the derive macro `Default` (in Nightly builds, run with -Z macro-backtrace for more info)
...
```

dhruvja: right right, then it means there is smtg wrong with the actions.

blasrodri:
Isn't there an anchor cli docker image we can use?

dhruvja:
> Isn't there an anchor cli docker image we can use?

anchor takes a very long time to build and run the tests. So we can just build it using sbf for now.

dhruvja:
> The CI should have failed, maybe its not working as expected.

The CI is not failing since the issue is already solved on master branch. So I think we are ready to merge.

.github/workflows/master.yml
dhruvja: Nothing is wrong with actions. Since the issue is resolved in master, the CI was passing.

mina86 (approved):

PR #90, opened 2023-11-14
common: use bytemuck to remove unsafe

mina86:
The unsafe is of course still there but now hidden in bytemuck
crate. The advantage is that bytemuck checks all the necessary
conditions when deriving TransparentWrapper rather than them only
being described in a comment.


dhruvja (approved):
LGTM

PR #91, opened 2023-11-14
change the string formatting in msg macro

dhruvja:
The `msg` macro which solana uses for logging is not the same as `println` macro i feel. Some of the formatting which we used to do in `println` macro like 
```
let x = 5;
println!("{x}"); 
```
Doesnt seem to work in `msg` macro. When used the above with msg macro, the output is -> `{x}`. So we have to specify empty braces and then pass the arguments outside the quotes. So this PR fixes that.


mina86 (approved):
Sorry about that.  That’s lame.  Also why I hate Rust formatting.

dhruvja:
> Sorry about that. That’s lame. Also why I hate Rust formatting.

Indeed it is LOL.

Issue #46, opened 2023-10-21
Use borsh for internal state encoding


mina86:
We’re currently using JSON to encode various state objects such as client state or consesus state. This is fine for now but for better performance we probably should eventually switch to borsh as demonstrated in [#44](https://github.com/ComposableFi/emulated-light-client/pull/44).

PR #95, opened 2023-11-16
solana-ibc: use Borsh for consensus state serialisation

mina86:
Closes: https://github.com/ComposableFi/emulated-light-client/issues/46

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ borsh = { version = "0.10.3", default-features = false }
 bytemuck = { version = "1.14", default-features = false }
 derive_more = "0.99.17"
 hex-literal = "0.4.1"
-ibc = { version = "0.47.0", default-features = false, features = ["serde", "borsh"] }
-ibc-proto = { version = "0.37.1", default-features = false, features = ["serde"] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -30,8 +30,8 @@ borsh = { version = "0.10.3", default-features = false }
 bytemuck = { version = "1.14", default-features = false }
 derive_more = "0.99.17"
 hex-literal = "0.4.1"
-ibc = { version = "0.47.0", default-features = false, features = ["serde", "borsh"] }
-ibc-proto = { version = "0.37.1", default-features = false, features = ["serde"] }
+ibc = { version = "0.47.0", default-features = false, features = ["borsh"] }
```
dhruvja, line 33: I think the `ibc::applications::transfer` requires the `serde` feature. So we have to keep the `serde` feature

mina86: Yes, you’re right. It’s weird that `cargo test` run didn’t fail compiling the code.

mina86: OK, actually, it’s probably because some dev dependency enables serde feature.

dhruvja: oh damn, got it

dhruvja (approved):

PR #97, opened 2023-11-16
solana-ibc: implement host_height method

mina86:
Properly implement host_height method to return Solana’s block height
(i.e. slot number) rather than using data from account.  The existing
implementation was just a quick hack to get some data going but is by
no means correct.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/host.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/host.rs
@@ -45,7 +45,7 @@ impl From<solana_program::clock::Clock> for Head {
 
 /// Error possible when fetching Solana’s clock.
 ///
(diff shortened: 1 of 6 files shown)


solana/solana-ibc/programs/solana-ibc/src/host.rs
```diff
@@ -0,0 +1,74 @@
...
+        Self {
+            height: clock.slot.into(),
+            timestamp: clock.unix_timestamp as u64,
+        }
+    }
+}
+
+/// Error possible when fetching Solana’s clock.
+///
+/// This is just a simple wrapper which offers trivial conversion no Solana and
```
blasrodri, line 48: Typo: on Solana

blasrodri (approved):

Issue #35, opened 2023-10-19
Investigate and implement possible restrictons on identifiers


mina86:
It would be mighty beneficial for us if we could restrict format of valid client, connection, port and channel identifiers.

For connection and channel we’re currently expecting those identifier to follow `<prefix>-<number>` format (where `<prefix>` is either `connection` or `channel` and `<number>` is `u32`). It would be great if we could impose similar restriction on client and port identifiers.

With such restrictions, we can parse the identifiers and convert them to plain `u32` value which can be compactly encoded in a trie path.

mina86:
Some thoughts I’ve been having about this over the weeks:
- For connection ids we can parse `connection-<number>` into u32.
- For channel ids we can parse `connection-<number>` into u32.
- For part ids we would need to impose some limits on them which go beyond what ibc-rs does.  I’m not really sure how flexible we’d need to be.
  * The easiest would of course be to impose `port-<number>` format and parse that into u32.
  * A more permissive option is to limit them to 8 (or maybe 12) alphanumeric characters.  Such strings can be padded with `/` and base64 decoded to 6 (or 9) bytes.  (Note that `PortId::transfer` constructor returns port id `transfer` which would fit that limitation; `PortId::default` returns `defaultPort` which is 11 letters long).
  * A more restrictive form of the above is to allow only lowercase letters and then 8 of them would decode using base32 to 5 bytes.
- For client ids it’s important to note that they are in `<client-type>-<number>` format and that the number is globally unique.  We could go one of two routes:
  * Whenever we have a new client save it’s identifier.  Then, parse `<client-type>-<number>` into u32 (or even u16) ignoring the client type.  With that index, check that the identifier matches what we’ve stored.  The check is needed so that we don’t accidentally treat `foo-1` and `bar-1` as the same client id.
  * Hard-code list of supported client types and then parse client id into u8 client type and u24 number.  An unknown client type will simply not parse.

mina86:
OK, we’ve launched so whatever we have now is what we’re going to be using. No point keeping this open.

PR #98, opened 2023-11-17
solana-ibc: map IBC client ids into u32 and refactor client state storage

mina86:
IBC client ids have `<client-type>-<counter>` format where counter is
a global sequential number.  Take advantage of that by converting
client ids into u32 including just the counter.

Since we’re effectively ignoring the client-type part of the id, keep
counter → client id map in private storage and use it to verify that
id we were given is the one we know about.

Furthermore, rather than storing client information in a map keep it
in a vector which is more compact and faster to index.  At the same
time, keep just a single vector for all per-client data rather than
having separate maps for each piece of information.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/trie_key.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/trie_key.rs
@@ -9,7 +9,7 @@ use ibc::core::ics24_host::path::{
 // those return the prefix without trailing `-` and we want constants which also
 // include that hyphen.
 use super::{CHANNEL_ID_PREFIX, CONNECTION_ID_PREFIX};
-use crate::storage::ClientIndex;
+use crate::storage::ClientIdx;
 
 /// A key used for indexing entries in the provable storage.
 ///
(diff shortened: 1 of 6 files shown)


solana/solana-ibc/programs/solana-ibc/src/trie_key.rs
```diff
@@ -232,6 +231,14 @@ impl AsComponent for ibc::core::ics24_host::identifier::ChannelId {
     }
 }
 
+impl AsComponent for ibc::Height {
+    fn key_len(&self) -> usize { 2 * 0_u64.key_len() }
+    fn append_into(&self, dest: &mut Vec<u8>) {
```
blasrodri, line 236: why not dest: &mut [u8] or `AsMut<[u8]>`?

mina86: Vec reallocates and tracks length of written data.  With `&mut [u8]` will have to preallocate maximum possible length and then extract length at the end.  Having said that, I will change it to `&mut &[u8]` once I convert all identifiers to integers.  Then we’ll be able to use ~30 byte fixed-size buffer for trie key.

dhruvja (approved):
LGTM

solana/solana-ibc/programs/solana-ibc/src/trie_key.rs
mina86: I’ve forgotten that we’re already calculating the length of the key and are doing proper preallocation, so yeah we could be using `&mut &[u8]` here even now.  Regardless, this is still planned once I convert all identifiers to integers.

PR #103, opened 2023-11-18
solana-ibc: refactor module structure

mina86:
Move trie_key inside of storage module and add storage::ids module
with ClientIndex type and other identifier related code.  The new
module will house other code related to converting IBC identifiers
into compact internal ones.


dhruvja (approved):
LGTM

PR #107, opened 2023-11-19
solana-ibc: remove commitment sequence sets from private storage

mina86:
The information stored in the packet commitment and acknowledgement
sequence sets is available in the trie.  Furthermore, the sets aren’t
used by the smart contract at all.  Drop them instead expecting the
off-chain code to rely on reading the trie.


dhruvja:
Just wondering, how we would get the packet commitments here. Can we search the trie?
https://github.com/ComposableFi/composable-ibc/blob/hyperspace-solana/hyperspace/solana/src/lib.rs#L431

mina86:
I’m now working on interface for the trie which lets scanning keys with given prefix. That would then give an option to get all the commitments from the trie.

dhruvja (approved):

PR #113, opened 2023-11-20
solana-ibc: store proof of client id together with client state

mina86:
As we parse client ids we strip out the client type.  This means
that ‘foo-42’ and ‘bar-42’ are internally represented as 42.  To
avoid confusion, we keep the client ids in private storage and
whenever we index client data by the number we check that the
full client id matches.

However, that information isn’t currently available in provable
storage (i.e. it’s not verifiable by light clients).

Change hash we’re storing under client state path to include the
client id.  This way, light clients will be able to verify that
indexes they’re extracting from the client id match the client id
they are given.

### DIFF
--- a/common/lib/src/hash.rs
+++ b/common/lib/src/hash.rs
@@ -45,7 +45,7 @@ impl CryptoHash {
     #[inline]
     pub fn digest(bytes: &[u8]) -> Self { Self::digestv(&[bytes]) }
 
-    /// Returns hash of consternation of given byte slices.
+    /// Returns hash of concatenation of given byte slices.
     ///
     /// This is morally equivalent to feeding all the slices into the builder
     /// one-by-one or concatenating them into a single buffer and hashing it in
--- a/common/lib/src/hash.rs
+++ b/common/lib/src/hash.rs
@@ -43,10 +43,31 @@ impl CryptoHash {
 
     /// Returns hash of given bytes.
     #[inline]
-    pub fn digest(bytes: &[u8]) -> Self {
-        let mut builder = Self::builder();
-        builder.update(bytes);
-        builder.build()
+    pub fn digest(bytes: &[u8]) -> Self { Self::digestv(&[bytes]) }
+
+    /// Returns hash of consternation of given byte slices.
+    ///
+    /// This is morally equivalent to feeding all the slices into the builder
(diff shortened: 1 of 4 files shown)


solana/solana-ibc/programs/solana-ibc/src/storage/trie_key.rs
```diff
@@ -74,6 +79,8 @@ impl TrieKey {
     }
 
     /// Constructs a new key for a connection end path.
+    ///
+    /// The hsah stored under the key is `hash(borsh(connection_end))`.
```
dhruvja, line 83:
```suggestion
    /// The hash stored under the key is `hash(borsh(connection_end))`.
```

solana/solana-ibc/programs/solana-ibc/src/storage/trie_key.rs
```diff
@@ -60,12 +60,17 @@ macro_rules! new_key_impl {
...
+    /// The hash stored under the key is `hash(borsh(client_id.as_str()) ||
+    /// borsh(client_state))`.
     pub fn for_client_state(client: ids::ClientIdx) -> Self {
         new_key_impl!(Tag::ClientState, client)
     }
 
     /// Constructs a new key for a consensus state path for client with given
     /// counter and specified height.
+    ///
+    /// The hsah stored under the key is `hash(borsh(consensus_state))`.
```
dhruvja, line 73:
```suggestion
    /// The hash stored under the key is `hash(borsh(consensus_state))`.
```

common/lib/src/hash.rs
```diff
@@ -43,10 +43,31 @@ impl CryptoHash {
 
     /// Returns hash of given bytes.
     #[inline]
-    pub fn digest(bytes: &[u8]) -> Self {
-        let mut builder = Self::builder();
-        builder.update(bytes);
-        builder.build()
+    pub fn digest(bytes: &[u8]) -> Self { Self::digestv(&[bytes]) }
+
+    /// Returns hash of consternation of given byte slices.
```
dhruvja, line 48:
```suggestion
    /// Returns hash of concatenation of given byte slices.
```

blasrodri (approved):

PR #114, opened 2023-11-20
solana-ibc: updated readme to remove running local validator

dhruvja:
We were running the local validator because we wanted increased
compute units (> 300k).  Since then we have added a new instruction
in tests to increase the compute units to 1M before each transaction.
This way we wouldn’t need to run a local validator.


dhruvja:
This can be only merged once #72 is merged since the change lies there. So i will keep it in draft till then.

PR #115, opened 2023-11-20
solana-ibc: introduce ChannelIdx type for channel identification

mina86:
Analogous to ConnectionIdx for connection identifiers, introduce
ChannelIdx type which represents a channel identifier.  It takes
advantage of our channels using `channel-<number>` format and
allows to store channel identifiers in more efficient way.


dhruvja (approved):
LGTM

PR #116, opened 2023-11-21
solana-ibc: restrict port identifier and use more compact representation

mina86:
Restrict port identifier to be 12 alphanumeric characters.  With that
restriction it’s now possible to decode the port using base64 encoding
to get 9 bytes.  Those bytes can then serve as an internal port
identifier used with the trie and maps in private storage.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/storage/ids.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/storage/ids.rs
@@ -1,4 +1,5 @@
 use anchor_lang::prelude::borsh;
+use base64::engine::{general_purpose, Engine};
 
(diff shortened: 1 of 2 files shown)


blasrodri (approved):
Can this also be extended to IBC go?

mina86:
This is purely on the implementation side so the same approach works
under ibc-go.  The spec describes this under [‘Identifier validation’
section of ICS
005](https://github.com/cosmos/ibc/blob/main/spec/core/ics-005-port-allocation/README.md#identifier-validation).

As far as I understand, by introducing this restriction we’re simply
restricting what ports can be bound but that’s just a matter for
whoever creates new connections to pick a supported port name.

PR #119, opened 2023-11-24
solana-ibc: make some modules public

dhruvja:
Many of the modules are private which makes them unable to use them
when the project is imported in the relayer. There are some interfaces
which are needed to be imported from this crate. So this PR makes the
required modules public.

Maybe we can split them into seperate crates in the future so that
they can be imported as needed.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -11,22 +11,22 @@ use ibc::core::ics24_host::identifier::PortId;
 use ibc::core::router::{Module, ModuleId, Router};
 use ibc::core::MsgEnvelope;
 
-const CHAIN_SEED: &[u8] = b"chain";
-const PACKET_SEED: &[u8] = b"packet";
-const SOLANA_IBC_STORAGE_SEED: &[u8] = b"private";
-const TRIE_SEED: &[u8] = b"trie";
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -122,7 +122,7 @@ pub mod solana_ibc {
         // Before anything else, try generating a new guest block.  However, if
         // that fails it’s not an error condition.  We do this at the beginning
         // of any request.
-        ctx.accounts.chain.maybe_generate_block(&provable, Some(host_head))?;
+        // ctx.accounts.chain.maybe_generate_block(&provable, Some(host_head))?;
```
mina86, line 125: Why is this a problem?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: While testing, we are not creating a chain, so whenever i call the deliver method, it panics because the chain is not initialized. We have to init the chain before we call the deliver method. But i need to know the config for initializing the chain though.

dhruvja: I will uncomment it for now so we can merge.

PR #121, opened 2023-11-24
sealable-trie: add bits::Owned::push and pop_back methods

mina86:
Also, for the time being, change bits::Owned::unshift to panic if
length of the slice ends up too long.  This is consistant with how Vec
handles things and in all places where unshift was used the caller
unwrapped the value so this isn’t such a terrible design for now.


common/sealable-trie/src/bits.rs
```diff
@@ -625,6 +719,12 @@ impl Owned {
             phantom: Default::default(),
         }
     }
+
+    /// Returns total number of underlying bits, i.e. bits in the slice plus the
+    /// offset.
+    fn underlying_bits_length(&self) -> usize {
+        usize::from(self.offset) + usize::from(self.length)
```
blasrodri, line 726: Can't overflow?

blasrodri (approved):

common/sealable-trie/src/bits.rs
mina86: Only on 16-bit platforms.

PR #122, opened 2023-11-24
make attributes of portChanelPk struct public

dhruvja:
Making them public so that the `channel_id` and `port_id` can be accessed. 

Should we add getter method or making them public is fine?

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/storage/ids.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/storage/ids.rs
@@ -1,4 +1,5 @@
 use anchor_lang::prelude::borsh;
+use ::ibc::core::ics24_host::identifier::ChannelId;
 
 use super::ibc;
 
@@ -125,8 +126,8 @@ impl TryFrom<&ibc::ConnectionId> for ConnectionIdx {
     borsh::BorshDeserialize,
 )]
 pub struct PortChannelPK {
-    pub port_id: ibc::PortId,
-    pub channel_idx: u32,
+    pub(super) port_id: ibc::PortId,
+    pub(super) channel_idx: u32,
 }
 
 impl PortChannelPK {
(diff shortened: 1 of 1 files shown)


mina86:
Why do you need it? Dosn’t conversion to `TrieKey` cover all use cases?

dhruvja:
> Why do you need it? Dosn’t conversion to `TrieKey` cover all use cases?

It does. I need it when i query the `channelEnds`.

dhruvja:
Do you think we should have getter methods so that it converts `ChannelIdx` to `ChannelId`. Some like this
```rs
    pub fn channel_id(&self) -> ibc::ChannelId {
        ChannelId::new(self.channel_idx.into())
    }

    pub fn port_id(&self) -> ibc::PortId {
        self.port_id.clone()
    }
```

mina86:
But if you’re querying channel ends what you need is `TrieKey::for_channel_end(&port_channel)`, no?

dhruvja:
> But if you’re querying channel ends what you need is `TrieKey::for_channel_end(&port_channel)`, no?

In one of the methods, it queries all the `channel_id` and `port_Id` set, which requires me to return the keys of the struct below. But since the `port_id` and `channel_id` are private, i cannot access it.
```rs
pub channel_ends: BTreeMap<ids::PortChannelPK, Serialised<ibc::ChannelEnd>>,
```

mina86:
OK, got it.  So what you need is to convert the PortChannelPK into `(ibc::PortId, ibc::ChannelId)`.

`channel_id` and `port_id` methods would be better here. `port_id` should return cloned id rather than a reference (this is because of https://github.com/ComposableFi/emulated-light-client/pull/116 which adds custom type for port id).

dhruvja:
> OK, got it. So what you need is to convert the PortChannelPK into `(ibc::PortId, ibc::ChannelId)`.
> 
> `channel_id` and `port_id` methods would be better here. `port_id` should return cloned id rather than a reference (this is because of #116 which adds custom type for port id).

right right, got it. I will make the changes.

mina86 (approved):

PR #127, opened 2023-11-24
solana-ibc: implement BorshDeserialize for Event

mina86:
So far Event lacked deserialisation derive because (to avoid memory
copies and allocations) it used references.  This worked great for
emitting the events but Borsh dosen’t support deserialising such
objects.

Replace CryptoHash and PubKey fields with owned values rather than
references and introduce CowBlock for block fields making Event enum
easily deserialisable.

To be honest I’m not happy with the current state of the type.  The
implementation is somewhat noisy with the necessity for CowBlock type,
but it should do for now.


solana/solana-ibc/programs/solana-ibc/src/events.rs
```diff
@@ -5,42 +5,77 @@ use lib::hash::CryptoHash;
...
+    Clone,
+    PartialEq,
+    Eq,
+    borsh::BorshSerialize,
+    borsh::BorshDeserialize,
+    derive_more::From,
+)]
 pub struct BlockFinalised<'a> {
     /// Hash of the block to which signature was added.
     pub block_hash: &'a CryptoHash,
```
dhruvja, line 81: We still have the reference to CryptoHash here. I think we should remove the reference here as well.
```suggestion
pub struct BlockFinalised {
    /// Hash of the block to which signature was added.
    pub block_hash:  CryptoHash,
```

solana/solana-ibc/programs/solana-ibc/src/events.rs
```diff
@@ -5,42 +5,77 @@ use lib::hash::CryptoHash;
...
+    borsh::BorshDeserialize,
+    derive_more::From,
+)]
 pub enum Event<'a> {
     IbcEvent(ibc::core::events::IbcEvent),
     Initialised(Initialised<'a>),
     NewBlock(NewBlock<'a>),
-    BlockSigned(BlockSigned<'a>),
+    BlockSigned(BlockSigned),
     BlockFinalised(BlockFinalised<'a>),
```
dhruvja, line 21:
```suggestion
    BlockFinalised(BlockFinalised),
```

solana/solana-ibc/programs/solana-ibc/src/events.rs
dhruvja: And also clone the hash value so that we dont take the reference here.
https://github.com/ComposableFi/emulated-light-client/blob/9a01809be249e231804f568cd67200d59548cc9d/solana/solana-ibc/programs/solana-ibc/src/chain.rs#L122

dhruvja (approved):

PR #128, opened 2023-11-25
solana-ibc: update ibc dependency

mina86:
ibc 0.48 restructured the modules hierarchy in quite a big way which
results in noisy diff around imports especially as ibc likes to use
quite long paths.  To make this somewhat more bearable, introduce ibc
module with re-exports of IBC types we care about.

Note that 0.48.0 fails to build in no_std on older rustc (which is
what is used with build-spf), thus go straight to 0.48.1 version.


dhruvja:
I hate that these guys bring breaking changes so often.

mina86:
I don’t really mind that much if there’s improvement in the API but this update was a bit annoying indeed.  Currently also blocked on https://github.com/cosmos/ibc-rs/issues/987

Farhad-Shabani:
Hey @dhruvja, @mina86, apologies for any hassle b/c of the recent release.
We strive to only change things when it's really essential for supporting a feature or a team's genuine needs. This `ibc` restructuring was aimed at unlocking diverse teams and use cases. I believe this is also much aligned with what [your team requested earlier](https://github.com/cosmos/ibc-rs/discussions/805). By the way, really appreciate your feedback!

mina86:
Well, it is what it is. I understand need for occasional breaking changes, but that won’t stop me complaining. ;)

dhruvja (approved):

PR #132, opened 2023-11-28
solana-ibc: added `send_packet` method

dhruvja:
Allows us to send packets given that we already have the client,
connection and channels set up.  This is different from
`send_transfer` and no transfer takes place in this method.  It
accepts a packet and stores it.

This method was added in IBC [v0.30.0](https://github.com/cosmos/ibc-rs/blob/8f2ddfefbeb7988bacc27dc6ab64ba7bcd537ce3/CHANGELOG.md?plain=1#L660)

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -1,20 +1,19 @@
 use blockchain::manager;
 
+use crate::ibc;
+
 /// Error returned when handling a request.
 // Note: When changing variants in the enum, try to preserve indexes of each
 // variant.  The position is translated into error code returned by Anchor and
 // keeping them consistent makes things easier.
-#[derive(strum::EnumDiscriminants, strum::IntoStaticStr)]
+#[derive(strum::EnumDiscriminants, strum::IntoStaticStr, derive_more::From)]
 #[strum_discriminants(repr(u32))]
 pub(crate) enum Error {
     /// Internal error which ‘should never happen’.
     Internal(&'static str),
 
-    /// Error handling an IBC deliver packet request.
-    RouterError(ibc::core::RouterError),
-
-    /// Error handling an IBC send packet request
-    ContextError(ibc::core::ContextError),
+    /// Error handling an IBC request.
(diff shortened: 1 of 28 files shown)


solana/solana-ibc/programs/solana-ibc/src/error.rs
```diff
@@ -10,9 +10,12 @@ pub(crate) enum Error {
     /// Internal error which ‘should never happen’.
     Internal(&'static str),
 
-    /// Error handling an IBC request.
+    /// Error handling an IBC deliver packet request.
     RouterError(ibc::core::RouterError),
 
+    /// Error handling an IBC send packet request
+    ContextError(ibc::core::ContextError),
+
```
mina86, line 18: Let’s wait for https://github.com/ComposableFi/emulated-light-client/pull/128 which will cover this.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -195,6 +196,32 @@ pub mod solana_ibc {
...
+            provable,
+            packets,
+            accounts: Vec::new(), // We are not doing any transfers so no point of having accounts
+            host_head,
+        });
+
+        ibc::core::send_packet(&mut store, packet)
+            .map_err(error::Error::ContextError)
+            .map_err(|err| error!((&err)))?;
+        Ok(())
```
mina86, line 223:
```suggestion
            .map_err(|err| error!((&err)))
```
should work here, no?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/error.rs
dhruvja: got it.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: Yes yes, the above would work.

dhruvja: Nvm, saw that its already been added to master with #128 .

mina86 (approved):

PR #133, opened 2023-11-28
solana-ibc: remove packets storage

dhruvja:
Storing packets on chain would easily pile up since they are just
added and not removed.  The previous delivered packets aren’t needed
in the smart contract so there is no point in storing them there.  The
packets are required by the relayer which can be fetched by querying
the logs.


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -146,15 +142,6 @@ pub mod solana_ibc {
                 .map_err(error::Error::RouterError)
```
mina86, line 142: There’s no longer need to clone message in line above.

dhruvja: right right, got it.

mina86 (approved):

PR #134, opened 2023-11-28
solana-ibc: use a proper timestamp in mock headers

dhruvja:
The timestamp default to `None` while setting up mock headers in the tests. Since they are not set ( `None` ), it fails while comparing the current timestamp. So using a value instead would solve it. The value has to be less than the current timestamp.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/tests.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/tests.rs
@@ -15,6 +15,9 @@ use anchor_lang::solana_program::instruction::AccountMeta;
 use anchor_lang::ToAccountMetas;
 use anchor_spl::associated_token::get_associated_token_address;
 use anyhow::Result;
+use ::ibc::core::client::types::Height;
+use ::ibc::primitives::Timestamp;
+use ibc_testkit::testapp::ibc::clients::mock::header::MockHeader;
 
 use crate::ibc::ClientStateCommon;
 use crate::storage::PrivateStorage;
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -3,6 +3,8 @@ use std::str::FromStr;
 use std::thread::sleep;
 use std::time::Duration;
 
+use ::ibc::core::client::types::Height;
+use ::ibc::primitives::Timestamp;
```
mina86, line 7: Both are available through `crtae::ibc` which is already imported. Use `ibc::Height` and `ibc::Timestamp`.

dhruvja: sure got it.

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -40,8 +40,12 @@ fn airdrop(client: &RpcClient, account: Pubkey, lamports: u64) -> Signature {
 
 fn create_mock_client_and_cs_state(
 ) -> (ibc::mock::MockClientState, ibc::mock::MockConsensusState) {
-    let mock_client_state = ibc::mock::MockClientState::new(Default::default());
-    let mock_cs_state = ibc::mock::MockConsensusState::new(Default::default());
+    let mock_header = ibc::mock::MockHeader {
+        height: ibc::Height::new(0, 1).unwrap(),
```
mina86, line 44:
```suggestion
        height: ibc::Height::min(0),
```
avoids unwrap

mina86 (approved):

PR #135, opened 2023-11-29
solana-ibc: introduce Owned::extend and Owned::truncate

mina86:
Add Owned::extend method which allows appending slices to the owned slice.

Do this through a generic concat::Concat trait and refactor other
concatenating methods to use a unified interface.  Specifically, this
removes unshift and push in favour of more flexible concat and a new
extend method.  Hopefully `Owned::concat(true, slice)` is more
descriptive than `Owned::unshift(true, slice)`.

With the new unified interfaces also comes more uniform error
handling.  Introduce concat::Error, concat::MisalignedSlice and
concat::SliceTooLong errors which are now consistently returned when
concatenation fails.  This covers length checking better and
eliminates panics if length of a concatenated slice is too large.

Furthermore, add Owned::truncate method which allows shortening of the
slice to specified length.  The method is akin to Vec::truncate and
has no effect if given length is longer than length of the slice.

### DIFF
--- a/common/sealable-trie/src/bits.rs
+++ b/common/sealable-trie/src/bits.rs
@@ -3,8 +3,10 @@ use core::fmt;
 
 use lib::u3::U3;
 
+pub mod concat;
 pub mod ext_key;
 
+pub use concat::MisalignedSlice;
 pub use ext_key::{Chunks, ExtKey};
 #[cfg(test)]
 use pretty_assertions::assert_eq;
@@ -23,7 +25,7 @@ pub struct Slice<'a> {
     pub(crate) offset: U3,
 
     /// Length of the slice in bits.
-    pub(crate) length: u16,
+    length: u16,
 
     /// The bytes to read the bits from.
     ///
@@ -31,7 +33,7 @@ pub struct Slice<'a> {
     /// unspecified and shouldn’t be read.
     // Invariant: if `length` is non-zero, `ptr` points at `offset + length`
     // valid bits; in other words, at `(offset + length + 7) / 8` valid bytes.
-    pub(crate) ptr: *const u8,
+    ptr: *const u8,
 
     phantom: core::marker::PhantomData<&'a [u8]>,
 }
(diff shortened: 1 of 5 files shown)


common/sealable-trie/src/bits.rs
```diff
@@ -440,39 +434,41 @@ impl<'a> Slice<'a> {
...
 }
 
+/// Calculates underlying bytes length of a slice with given offset and length.
+#[inline]
+fn bytes_len(offset: U3, length: u16) -> usize {
+    // We need to special-case zero length to make sure that in situation of
+    // non-zero offset and zero length we return an empty slice.
+    match length {
+        0 => 0,
+        _ => ((u32::from(offset) + u32::from(length) + 7) / 8) as usize,
```
blasrodri, line 468: is the numerator always a multiple of 8? does it matter at all?

common/sealable-trie/src/bits/concat.rs
```diff
@@ -0,0 +1,455 @@
+use alloc::vec::Vec;
+
+use lib::u3::U3;
+
+use super::{Owned, Slice};
+
+
+/// Trying to concatenate slices which result in slice whose size is too large.
```
blasrodri, line 8: why do we have to implement this? no existing library can give us these features?

common/sealable-trie/src/bits.rs
mina86: No, numerator can be arbitrary number. This divides by eight with rounding up to calculate number of bytes needed for given number of bits.  That is, if `offset + length` is say 10 (bits) that occupies 2 (bytes).

common/sealable-trie/src/bits/concat.rs
mina86: Honestly we probably can use [bitvec](https://docs.rs/bitvec/latest/bitvec/index.html).  Rewriting the code is just something I haven’t gotten to doing.  The bits module grew organically to fit the needs of the trie where each change was relatively simple.  Because of that each time just adding another small feature seemed like less work than rewriting everything to use bitvec.  But it’s definitely something I need to look into at some point and definitely before prod release.

blasrodri (approved):

PR #136, opened 2023-11-29
solana-ibc: introduce Trie::get_subtrie call

mina86:
The call returns all keys within a subtrie with given prefix.  The
feature is needed by the relayer to scan the trie for pending packets.


dhruvja:
@blasrodri are we good with this?

PR #138, opened 2023-11-30
trie-ids: extract trie key and ids handling to separate crate

mina86:
To reason about data in the provable storage, i.e. the trie, code
needs access to the TrieKey type.  At the moment it’s defined within
the solana-ibc crate which means that to use it external crates need
to pull in a large dependency including all of Anchor and Solana SDK.

This is acceptable by relayer which needs to have access to other
types defined in the Solana contract, but the TrieKey is also needed
by the light client for which pulling all of solana-ibc would be at
the very least questionable.

To address this, move TrieKey and related type definitions into its
own crate.

I’ve contemplated moving it inside of blockchain crate but concluded
that it only deals with the block governance without carrying what’s
stored in the trie so keeping TrieKey there wouldn’t fit the crate.

With this, add a few simple tests for the types and fix borsh
deserialisation issue around PortKey where an invalid port key could
be constructed through a crafted serialised representation.  This is
now not possible.


mina86:
@dhruvja, with relayer using older IBC do we need here support for older IBC as well?

dhruvja:
> @dhruvja, with relayer using older IBC do we need here support for older IBC as well?

the relayer uses new IBC now so i dont think we need to support older IBC.

PR #140, opened 2023-12-04
solana-ibc: use named optional accounts

### LINKED ISSUE #104: Use Optional Accounts
Right now, we are using `remaining_accounts` for doing transfer. `remaining_accounts` are just an array of accounts which are not validated by `anchor`. There are chances of passing some wrong accounts and we dont actually know which accounts to pass. So using optional accounts would give us the validation by `anchor` and also allow us to do any initialization too. 

It might be something like below
```rs
 #[account(init_if_needed, payer = sender, seeds = [base_denom.as_bytes()], bump, mint::decimals = 6, mint::authority = mint_authority)]
 token_mint: Option<Account<'info, Mint>>,
```

Having this would allow us to initialize the account if not done previously and would also allow us to know which accounts we can expect.

The current implementation is good for now. We can switch to this when we have some time in future.

dhruvja:
Closes #104


dhruvja:
Not ready to merge because currently blocked by https://github.com/coral-xyz/anchor/pull/2101#issuecomment-1837941414 . So i will keep in draft until i find a solution for it.

dhruvja:
I added another method `delivery_init_escrow` where the accounts ( token mint and escrow ) needed for transfer are initialized incase they are not. It is similar to the implemented `mock_init_escrow`. This is done so that we dont run out of stack since initializing accounts take quite a lot of stack space.

dhruvja:
@mina86

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -115,11 +116,46 @@ pub mod solana_ibc {
         ctx.accounts.chain.set_stake((*ctx.accounts.sender.key).into(), amount)
     }
 
+    /// Called to set up escrow and mint accounts for given channel and denom.
+    #[allow(unused_variables)]
+    pub fn deliver_init_escrow<'a, 'info>(
+        ctx: Context<'a, 'a, 'a, 'info, DeliverInitEscrow<'info>>,
+        port_id: ibc::PortId,
+        channel_id_on_b: ibc::ChannelId,
+        base_denom: String,
```
mina86, line 125: Perhaps just `init_escrow`?

```suggestion
    pub fn init_escrow<'a, 'info>(
        _ctx: Context<'a, 'a, 'a, 'info, DeliverInitEscrow<'info>>,
        _port_id: ibc::PortId,
        _channel_id_on_b: ibc::ChannelId,
        _base_denom: String,
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -115,11 +116,46 @@ pub mod solana_ibc {
...
+    ) -> Result<()> {
+        Ok(())
```
mina86, line 127: Could you add a comment explaining why the comment is empty? I.e. that all work is done by Anchor when accounts are initialised.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -344,9 +408,31 @@ pub struct Deliver<'info> {
     trie: UncheckedAccount<'info>,
 
     /// The guest blockchain data.
-    #[account(mut, seeds = [CHAIN_SEED], bump)]
-    chain: Account<'info, chain::ChainData>,
+    #[account(mut, seeds = [CHAIN_SEED],
+              bump)]
+    chain: Box<Account<'info, chain::ChainData>>,
+    #[account(mut, seeds = [MINT_ESCROW_SEED],
+        bump)]
```
mina86, line 415:
```suggestion
    #[account(mut, seeds = [CHAIN_SEED], bump)]
    chain: Account<'info, chain::ChainData>,
    #[account(mut, seeds = [MINT_ESCROW_SEED], bump)]
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -168,7 +204,7 @@ pub mod solana_ibc {
         timeout_height: ibc::TimeoutHeight,
         timeout_timestamp: ibc::Timestamp,
     ) -> Result<()> {
-        let mut store = storage::from_ctx!(ctx);
+        let mut store = crate::storage::from_ctx!(ctx);
```
mina86, line 207:
```suggestion
        let mut store = storage::from_ctx!(ctx);
```

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -352,12 +352,34 @@ pub fn get_provable_from<'a, 'info>(
     get(info).map_err(|err| err.with_account_name("trie"))
 }
 
+/// Used for finding the account info from the keys.
+/// Useful for finding the token mint on the source chain which cannot be derived from the denom.
+/// Would also save us some compute units to find authority and other accounts which used to be found by deriving from the seeds.
```
mina86, line 357: Could you wrap at 80 columns? And separate first line by an empty comment so it shows as short description in rendered docs.

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -352,12 +352,34 @@ pub fn get_provable_from<'a, 'info>(
...
+pub enum TransferAccountNames {
+    /// Should be a signer
+    Sender,
+    SenderTokenAccount,
+    ReceiverTokenAccount,
+    TokenMint,
+    EscrowAccount,
+    MintAuthority,
+    TokenProgram,
+}
```
mina86, line 375: I’m not sold on this. Why not make it a struct with those fields? Currently, how do you access TokenMint for example? You need to iterate over the vector?

```rust
pub struct TransferAccounts<'a, 'info> {
    sender: Option<&'a AccountInfo<'info>>,
    sender_oken_account: Option<&'a AccountInfo<'info>>,
    receiver_token_account: Option<&'a AccountInfo<'info>>,
    token_mint: Option<&'a AccountInfo<'info>>,
    escrow_account: Option<&'a AccountInfo<'info>>,
    mint_authority: Option<&'a AccountInfo<'info>>,
    token_program: Option<&'a AccountInfo<'info>>,
}
```

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -431,7 +452,7 @@ macro_rules! from_ctx {
             private,
             provable,
             chain,
-            accounts,
+            accounts: &[],
         })
     }}
 }
```
mina86, line 458: What you wanna do here is add other invocations of the macro.
Something like:

```rust
macro_rules! from_ctx {
    ($ctx:expr) => {
        from_ctx!($ctx, Default::default())
    };
    ($ctx:expr with accounts) => {
        let account = {
            // put body of to_token_accounts here
        };
        from_ctx!($ctx, accounts)
    };

    ($ctx:expr, $accounts:expr) => {{
        let provable = storage::get_provable_from(&$ctx.accounts.trie)?;
        let chain = &mut $ctx.accounts.chain;

        // Before anything else, try generating a new guest block.  However, if
        // that fails it’s not an error condition.  We do this at the beginning
        // of any request.
        chain.maybe_generate_block(&provable)?;

        storage::IbcStorage::new(storage::IbcStorageInner {
            private: &mut $ctx.accounts.storage,
            provable,
            chain,
            accounts: $accounts,
        })
    }}
}
```

and then you should be able to do:

```
        let mut store = storage::from_ctx!(ctx with accounts);
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -115,11 +116,46 @@ pub mod solana_ibc {
...
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    #[allow(unused_variables)]
     pub fn deliver<'a, 'info>(
         ctx: Context<'a, 'a, 'a, 'info, Deliver<'info>>,
+        port_id: Option<ibc::PortId>,
+        channel_id_on_b: Option<ibc::ChannelId>,
+        base_denom: Option<String>,
```
mina86, line 135: So the reason for those arguments is that we need them for seeds? And their value, if set, is the same as one specified somewhere inside of message?

dhruvja: So we either check validate the seeds here or we do in send validate. If we do in send validate, there would not be any need of these params which would be better IMO.

mina86: We need those values in `seeds` annotation though, no?

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: Well I'm wondering why I chose this method. Struct is only better. I think my initial thought was to iterate over the keys but if they are named, no point in doing so. I will switch to struct.

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: Sure

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: Oh ya sure ill add

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: It's not necessary, we can skip it. It's only for validation which we can do it in send validate.

mina86: SGTM

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: nice, makes sense. Could do that.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -448,6 +530,39 @@ pub struct SendPacket<'info> {
...
 }
 
+impl<'a> Deliver<'a> {
+    fn to_transfer_accounts(&self) -> TransferAccounts<'a> {
+        TransferAccounts {
+            sender: Some(self.sender.as_ref().to_account_info()),
+            receiver: self
+                .receiver
+                .clone()
+                .map(|account| account.to_account_info()),
```
mina86, line 540: Alternatively:
```suggestion
                .map(ToAccountInfo::to_account_info),
```
Matter of taste I suppose.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -115,11 +116,48 @@ pub mod solana_ibc {
         ctx.accounts.chain.set_stake((*ctx.accounts.sender.key).into(), amount)
     }
 
+    /// Called to set up escrow and mint accounts for given channel and denom.
+    /// The body of this method is empty since this method is only called for initializing the accounts.
+    /// Anchor initializes the accounts mentioned before the body of the method is called.
```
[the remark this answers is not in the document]
dhruvja: right right

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: But i was able to do the above for accounts which are not boxed ( `receiver` ).

dhruvja: Nvm it worked with `deref` instead.

PR #141, opened 2023-12-04
sealable-trie: prohibit storing values at key prefixes

mina86:
Due to the nature of sealable trie, storing values at a prefix of an
existing key isn’t supported.  To understand why, consider the
following sequence of operations:

* set 0000 → foo
* set 0001 → bar
* seal 0000
* seal 0001
* set 000 → baz

After the second seal operation, node 000 is sealed as well.  This is
what makes sealing a worthwhile pursue allowing us to prune subtries
as the leafs are sealed.

But if 000 is sealed than setting 000 → baz cannot happen even though
the key wasn’t explicitly sealed.  However, if that operation was
performed earlier, it would succeed with the key being able to be
modified.

Similarly, it’s already been codded that sealing a key seals the
entire subtrie.

All of this is somewhat surprising and inconsistent.  To avoid all
that confusion, simply prohibit setting values for prefixes of
existing keys.

### DIFF
--- a/common/sealable-trie/src/bits/ext_key.rs
+++ b/common/sealable-trie/src/bits/ext_key.rs
@@ -61,9 +61,6 @@ impl<'a> ExtKey<'a> {
     /// Returns the length of relevant portion of the buffer.  For example, if
     /// slice’s length is say 20 bits with zero offset returns five (two bytes
     /// for the encoded length and three bytes for the 20 bits).
-    ///
-    /// Returns `None` if the slice is empty or too long and won’t fit in the
-    /// destination buffer.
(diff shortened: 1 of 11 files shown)


blasrodri (approved):
can you add the same test you described on the PR?
- add 0000 -> foo
- add 0001 -> bar
- seal them
- add 000 -> baz

mina86:
This exact scenario is impossible because the API doesn’t actually provide bit-granularity for the keys. All keys cover full bytes. Nonetheless, I’ve added test where all 0x00?? keys are set and then 0x00 key is manipulated.

PR #143, opened 2023-12-07
solana-ibc: implement guest blockchain’s consensus state

mina86:
Define ConsensusState type together with protocol message and
implement the host_consensus_state method to return consensus state of
the guest blockchain at head height.  Getting consensus state of past
blocks isn’t supported at this moment and is still left to be done.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -27,6 +27,11 @@ jobs:
           toolchain: nightly
           components: clippy rustfmt miri
 
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Check formatting
         run: cargo fmt --all --check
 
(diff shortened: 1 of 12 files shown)


dhruvja:
Looks like we are having stack overflow issues in `mock_deliver`. I tried to box a the members of `ChainInner` and it seemed go away, then we are getting `memory_allocation_failed` while storing in private storage.

mina86:
Yes, I’m fighting with that as well.

Rather than boxing members of ChainInner boxing the entire ChainInner might be a better option (but memory allocation still remains an issue).

dhruvja:
The `chainData` seems to be taking a lot of heap memory. For implementing `zero_copy`, the data structure needs to implement `Copy` .

mina86:
What we probably need is an account where part of it is zero-copy and part is serialised. A bit like with trie where header is parsed and the rest of the account is accessed directly. I’m not sure at the moment if there’s easy support for something like that in Anchor.

mina86:
PTAL. Works with history excluded. This at least gives us something to test with.

mina86:
Yes, for the time being we can live without history. History is only used when establishing connections and only because head of the chin changes as connection is being established. For testing I think we should be able to do everything within a single block; or at the worst two blocks.

PR #144, opened 2023-12-08
trie-ids: make `Tag` and related methods and traits public to create sub trie keys

dhruvja:
Since we can fetch from trie using prefix of the key which is used to
fetch all the packet commitments, acks and recv, we need to make `Tag`
enum and the methods used to create new key as public so that we can
create sub trie keys.


solana/solana-ibc/programs/solana-ibc/src/tests.rs
mina86, line 1: Probably remove that change since it’s only comments.

PR #145, opened 2023-12-08
solana-ibc: fix `anchor build` and add it to CI

mina86:
Anchor doesn’t like when array sizes aren’t integer literals.  This is
dumb but there’s not much we can do other than change argument type
from a somewhat descriptive `[u8; ed25519::Signature::LENGTH]` to
a somewhat cryptic `[u8; 64]`.  This fixes `anchor build` run.

With that, add Anchor Build to CI so it’s tested on each PR.  While
doing that, rearrange CI jobs a little so that their run time is more
balanced.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -50,6 +50,23 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-features -- -D warnings
 
+      - name: Miri tests
+        run: cargo miri test -- -Z unstable-options --report-time --skip ::anchor
+
+  anchor-build:
+    name: Anchor Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        id: install-rust
+        uses: dtolnay/rust-toolchain@master
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/master.yml
```diff
@@ -38,19 +50,74 @@ jobs:
...
+          avm install $ANCHOR_VERSION
+          avm use $ANCHOR_VERSION
+
+      - name: Anchor Build
+        run: anchor build
+
+      - name: Anchor Build (with mocks)
+        run: anchor build -- --features=mocks
+
+  solana-build:
```
dhruvja, line 106: I dont think we have to build we need this, because `anchor build` runs `cargo build-sbf` under the hood anyways.

mina86: I wondered about that. What stopped me though is that `anchor build` takes a bit over five minutes while `cargo build-sbf` takes a bit over six minutes. Though it’s possible that anchor is smarter about what it builds with sbf I suppose.

dhruvja: I thinks anchor optimizes somehow. https://github.com/coral-xyz/anchor/blob/master/cli/src/config.rs#L425

dhruvja (approved):

PR #147, opened 2023-12-10
solana-ibc: initialise chain in Anchor test

mina86:
Most notably, add chain initialisation to the Anchor test. During chain initialization, we also initialize all the storage accounts ( private storage, trie and chain data ) since its only called once in the lifetime of the program. Hence the use `inif_if_needed` for storage accounts serve no purpose thus save us some from `re-initialization` attacks. While at
it, change some logging in the test so it’s easier to read the logs.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/Cargo.toml
+++ b/solana/solana-ibc/programs/solana-ibc/Cargo.toml
@@ -42,3 +42,5 @@ anchor-client.workspace = true
 anyhow.workspace = true
 ibc-testkit.workspace = true
 insta.workspace = true
+
+lib = { workspace = true, features = ["test_utils"] }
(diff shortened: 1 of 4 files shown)


dhruvja (approved):
Added initialization of accounts to `Chain` and removed `init_if_needed` for data accounts everywhere. @mina86

mina86:
Does it need to be `init` in other structs? Should `mut` suffice?

dhruvja:
> Does it need to be `init` in other structs? Should `mut` suffice?

Its only init in `Chain` struct. Everywhere else it should be mut. making it `init` would panic since the program would say that the `account already exists`.

mina86:
Yeah, I’ve just noticed after making my comment.

PR #148, opened 2023-12-10
solana-ibc: introduce separate mock_init_escrow instruction

mina86:
Split out mint and escrow initialisation out of mock_deliver and into
mock_init_escrow instruction.  This is done so that the former has
fewer things to initialise and we don’t run into stack or heap memory
issues.

Also, sneak in addition of chain field to the MockDeliver to
demonstrate that this indeed achieves desired result.  Previously,
adding the field into the struct would lead to stack frame overflow.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -28,7 +28,7 @@ pub mod events;
 mod execution_context;
 mod host;
 mod ibc;
-#[cfg(feature = "mocks")]
+#[cfg_attr(not(feature = "mocks"), path = "no-mocks.rs")]
 mod mocks;
 pub mod storage;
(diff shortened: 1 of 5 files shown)


dhruvja:
The thing is we can get away in the mocks, but eventually we will have to initialize `token_mint` and `escrow_account` in the `deliver` method. So we have to find other method or do the same for `deliver` like above though i think it can get tricky.

mina86:
Yes, I think we will need to do the same for deliver. Perhaps it’s better that way anyway since we can make this a privileged operation and thus control what tokens the bridge supports.

dhruvja:
> Yes, I think we will need to do the same for deliver. Perhaps it’s better that way anyway since we can make this a privileged operation and thus control what tokens the bridge supports.

right right, makes sense. Alright then.

dhruvja (approved):

PR #159, opened 2023-12-15
solana-ibc: remove host module

mina86:
Idea behind host module was that it would grow with Solana-specific
helper code.  The module started shrinking instead.  Since it’s only
type is used in a single module move the code there.  With that
simplify it and better document what we use as height in Solana.


solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -188,25 +188,25 @@ impl ChainInner {
...
-        if self.last_check_height == host_head.height {
+        if self.last_check_height == host_height {
             return if force {
                 Err(Error::GenerationAlreadyAttempted.into())
             } else {
                 Ok(())
             };
         }
-        self.last_check_height = host_head.height;
+        self.last_check_height = host_height;
```
dhruvja, line 206: As long as change in slot indicates a new block, it should be fine

mina86: It does. If a slot changes, next leader gets to generate next block.

dhruvja (approved):

PR #162, opened 2023-12-19
blockchain: separate BlockHeader out of Block

mina86:
Separate BlockHeader out of Block with all data except for the next
epoch.  Instead, header only stores the next epoch commitment
(i.e. hash of the serialised epoch).

Reason for doing this is that serialised epoch is noticeably larger
than just the block.  Meanwhile, there are operations which one may
want to perform without having the next epoch provided.

For instance, proving past block hash can be done by sending a chain
of block headers.  In old code, if that chain crosses epoch boundary,
the epoch would have to be included.  Now, only the epoch hash needs
to be provided.

Part of the motivation is also Solana where instructions are quite
limited in size with a single transaction having to fit in
roughtly 1.2 KB.  If an epoch defines 20 validators, it’s serialised
size is already almost 1 KB.


common/blockchain/src/block.rs
```diff
@@ -42,10 +40,81 @@ pub struct Block<PK> {
     /// very next block will use current’s block hash as `epoch_id`.
     pub epoch_id: CryptoHash,
 
+    /// If present, hash of the serialised epoch *the next* block will belong
```
blasrodri, line 43: when would it be present, and when wouldn't it?

mina86: added more docs

blasrodri (approved):

PR #164, opened 2023-12-20
blockchain: implement proof handling in a way IBC expects

mina86:
Introduce blockchain::ibc_state::proof module with code for generating
and validating proofs in the format that IBC expects.  The code uses
IBC data types and include all the information to go from consensus
commitment root to verifying particular value is stored in the trie.


common/blockchain/src/ibc_state/proof.rs
```diff
@@ -0,0 +1,435 @@
...
+    pub proof: ibc::CommitmentProofBytes,
+    /// Commitment root.
+    pub root: ibc::CommitmentRoot,
+    /// Value stored at the path (if it exists).
+    pub value: Option<CryptoHash>,
+}
+
+impl IbcProof {
+    /// Returns commitment prefix to use during verification.
+    pub fn prefix(&self) -> ibc::CommitmentPrefix { Default::default() }
```
blasrodri, line 25: we have no prefix?

common/blockchain/src/ibc_state/proof.rs
```diff
@@ -0,0 +1,435 @@
...
+///
+/// This addition is necessary because sequence numbers are stored together
+/// within a single trie value.  For example, proving the next receive
+/// sequence is `4` requires proving `be(7), be(4), be(5), be(0)].  For
+/// verifier to know what value it checks, it needs to be provided all of
+/// the sequence numbers.
+///
+/// (Note that Borsh uses little endian to encode integers so the sequence
+/// numbers cannot be simply borsh deserialised.)
+pub fn generate<A: sealable_trie::Allocator>(
```
blasrodri, line 72: why not `generate_proof`

common/blockchain/src/ibc_state/proof.rs
mina86: Not that I can think of. We’re storing everything at the root of the trie.

common/blockchain/src/ibc_state/proof.rs
mina86: Well, the full path to to the function is `ibc_state::proof::generate` so `proof` is already there.

blasrodri (approved):

PR #165, opened 2023-12-23
added method to form message using chunks

dhruvja:
Since instructions to solana have a limit of 1232 bytes, we cannot send messages which exceed the limit. Since the update headers and other messages can easily exceed the limit, we have to split them into chunks and add it to an account so that it can be accessed later in the `deliver` method.

Once the data in the account in executed by the `deliver`, the account data ( which stored the msg chunks) would be cleared.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -145,6 +145,18 @@ pub mod solana_ibc {
             .map_err(move |err| error!((&err)))
     }
 
+    pub fn deliver_with_chunks<'a, 'info>(
+        ctx: Context<'a, 'a, 'a, 'info, DeliverWithChunks<'info>>,
+    ) -> Result<()> {
+        let msg_chunks = &ctx.accounts.msg_chunks;
+        let mut store = storage::from_ctx!(ctx, with accounts);
+        let mut router = store.clone();
+        let message = ibc::MsgEnvelope::try_from_slice(&msg_chunks.msg).unwrap();
+        ::ibc::core::entrypoint::dispatch(&mut store, &mut router, message)
+            .map_err(error::Error::ContextError)
+            .map_err(move |err| error!((&err)))
+    } 
+
     /// Called to set up escrow and mint accounts for given channel and denom.
     /// Panics if called without `mocks` feature.
     pub fn mock_init_escrow<'a, 'info>(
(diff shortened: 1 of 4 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -144,6 +146,23 @@ pub mod solana_ibc {
...
+        ctx: Context<'a, 'a, 'a, 'info, DeliverWithChunks<'info>>,
+    ) -> Result<()> {
+        let msg_chunks = &ctx.accounts.msg_chunks;
+        let cloned_msg_chunks: &mut MsgChunks = &mut msg_chunks.clone();
+        let mut store = storage::from_ctx!(ctx, with accounts);
+        let mut router = store.clone();
+
+        let message = ibc::MsgEnvelope::try_from(ibc::Any::from(
+            cloned_msg_chunks.clone(),
+        ))
```
mina86, line 159: Why do we clone msg_chunks twice? Why is `cloned_msg_chunks` an exclusive reference?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -505,6 +588,19 @@ pub struct SendPacket<'info> {
     system_program: Program<'info, System>,
 }
 
+#[derive(Accounts)]
+pub struct FormMessageChunks<'info> {
+    #[account(mut)]
+    sender: Signer<'info>,
+
+    #[account(init_if_needed, payer = sender, seeds = [MSG_CHUNKS], bump, space = 10240)]
+    pub msg_chunks: Account<'info, storage::MsgChunks>,
```
mina86, line 597: Two things.

First of all, do we need this to be a PDA? I was actually wondering if it wouldn’t be better if this was just an account owned by the sender and then they can manage it in any way they want.  All we need here is a writeable account.

Second of all, if this is a PDA it *must* include sender in the seeds.  Without that, users will overwrite each others calls.

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -379,6 +379,40 @@ pub(crate) struct IbcStorageInner<'a, 'b> {
...
+/// Struct containing message chunks
+///
+/// The struct consists of chunks of message data having the first 4 bytes
+/// indicating the total size of the message
+#[account]
+#[derive(Debug)]
+pub struct MsgChunks {
+    pub type_url: String,
+    pub value: Vec<u8>,
+}
```
mina86, line 391: This really should be zero copy or unchecked account IMO.  Furthermore, rather than giving it a structure with type url and value, it’s better to treat it as a length-prefixed sequence of bytes.  Those can than be decoded as proto to get Any message.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: yes i was thinking about it too. I was also looking if we could create an account with data outside of smart contract. Couldnt find anything as now. But we dont need PDA, we could use the sender's account too. Once the `deliver_chunks` is called, we can also close the account so that we can return the rent.

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: the value is length prefixed but we would still need `type_url` right?

mina86: No. The value would be an Any proto message so it would include type url.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: its a reference because i needed msg_chunks as `MsgChunks` so that i could convert it. But i think we can ignore it as we are doing the deserializing stuff in the `form_msg_chunks` method itself. So we would just be fetching the `MsgEnvelope` from the account and hence no conversions would be required in the `deliver_chunks` method.

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: Oh i see, that makes sense.

dhruvja:
Closing this since we would be using https://github.com/ComposableFi/emulated-light-client/pull/171 .

PR #167, opened 2023-12-29
solana-ibc: remove unneccesary Ok in validation and execution context

dhruvja:
Replacing explicit `Ok(())` with the `Result` provided by the methods.


mina86 (approved):
> @mina86 do we want to return an error saying the key was not found while deleting the key from the trie?
> For example: https://github.com/ComposableFi/emulated-light-client/blob/977366d2a848bc8b0cb95ca9312d1ffa0d620157/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs#L76

I think so.  ibc-rs shouldn’t call the delete if it didn’t fetch the consensus previously.

PR #169, opened 2023-12-29
converting timestamp into nanoseconds while conversion in `host_timestamp`

dhruvja:
The timestamp which we get from the `chain` head is in seconds but while converting it to ibc timestamp, we have to convert it into nanoseconds. So multiplying the timestamp by 10^9 to get it in nanoseconds.


mina86:
https://github.com/ComposableFi/emulated-light-client/pull/172 makes this obsolete.

PR #171, opened 2023-12-29
write-account: add Solana program for writing data into an account

mina86:


mina86:
Haven’t testing it at all but this is the kind of contract I had in mind.

blasrodri (approved):

PR #173, opened 2024-01-02
Add Cargo.lock

mina86:
We’re having issues where pulling in new versions of dependencies
leads to build failures.  For example, the latest instance is ahash
v0.8.7 which is not compatible with Solana build process and results
in the following error:

    error[E0658]: use of unstable library feature 'build_hasher_simple_hash_one'
       --> src/random_state.rs:463:5
        |
    463 | /     fn hash_one<T: Hash>(&self, x: T) -> u64 {
    464 | |         RandomState::hash_one(self, x)
    465 | |     }
        | |_____^
        |

To avoid those problems, include Cargo.lock in the repository.


dhruvja (approved):
LGTM

PR #174, opened 2024-01-02
solana-ibc: send transfer implementation

dhruvja:


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -505,6 +525,47 @@ pub struct SendPacket<'info> {
...
+#[instruction(port_id: ibc::PortId, channel_id_on_b: ibc::ChannelId, base_denom: String)]
+pub struct SendTransfer<'info> {
+    #[account(mut)]
+    sender: Signer<'info>,
+
+    receiver: Option<AccountInfo<'info>>,
+
+    /// The account holding private IBC storage.
+    #[account(mut,seeds = [SOLANA_IBC_STORAGE_SEED],
+              bump)]
```
mina86, line 538:
```suggestion
    #[account(mut,seeds = [SOLANA_IBC_STORAGE_SEED], bump)]
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -505,6 +525,47 @@ pub struct SendPacket<'info> {
...
+    storage: Account<'info, storage::PrivateStorage>,
+
+    /// The account holding provable IBC storage, i.e. the trie.
+    ///
+    /// CHECK: Account’s owner is checked by [`storage::get_provable_from`]
+    /// function.
+    #[account(mut, seeds = [TRIE_SEED],
+              bump)]
```
mina86, line 546:
```suggestion
    #[account(mut, seeds = [TRIE_SEED], bump)]
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -505,6 +525,47 @@ pub struct SendPacket<'info> {
...
+    /// The guest blockchain data.
+    #[account(mut, seeds = [CHAIN_SEED], bump)]
+    chain: Box<Account<'info, chain::ChainData>>,
+    #[account(mut, seeds = [MINT_ESCROW_SEED], bump)]
+    /// CHECK:
+    mint_authority: Option<UncheckedAccount<'info>>,
+    #[account(mut)]
+    token_mint: Option<Box<Account<'info, Mint>>>,
+    #[account(init_if_needed, payer = sender, seeds = [
+        port_id.as_bytes(), channel_id_on_b.as_bytes(), base_denom[..18].as_bytes(), base_denom[18..].as_bytes()
```
mina86, line 558: Why is base denom twice? Why only first 18 bytes? What if byse denom is shorter than 18 bytes?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -263,6 +263,26 @@ pub mod solana_ibc {
...
+        ctx: Context<SendTransfer>,
+        port_id: ibc::PortId,
+        channel_id_on_b: ibc::ChannelId,
+        base_denom: String,
+        msg: ibc::MsgTransfer,
+    ) -> Result<()> {
+        let mut store = storage::from_ctx!(ctx, with accounts);
+        let mut token_ctx = store.clone();
+
+        ::ibc::apps::transfer::handler::send_transfer(
```
mina86, line 278: `ibc::aps::transfer` should work (without the `::` in front. Here and throughout the PR.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -103,6 +103,7 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
 
         let (sender, receiver, authority) =
             if matches!(from, AccountId::Escrow(_)) {
+                msg!("not here");
```
mina86, line 106: That’s a rather peculiar message.

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: when we send the token from solana, the denom would be the token mint which is always 32 bytes. And on the seeds, it has a limit on the length because when i tried to add `base_denom` without splitting, it was not able to derive the PDA. I wanted to reduce the length so i took an arbitrary value of 18. But we can use 16 since thats the half of the denom.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: cool

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: haha, didnt want that to get logged in `send_transfer` method lol. I will remove it since it was only for testing.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
mina86: Got it. It shouldn’t matter since seeds are concatenated but perhaps let’s split at 32 and add a comment describing why we’re doing it?

dhruvja: sure sure ill add

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -552,8 +552,9 @@ pub struct SendTransfer<'info> {
     mint_authority: Option<UncheckedAccount<'info>>,
     #[account(mut)]
     token_mint: Option<Box<Account<'info, Mint>>>,
+    /// Splitting the denom since we can have a max of only 32 bytes
```
mina86, line 555:
```suggestion
    // Splitting `base_denom` since each seed can be at most 32 byte long.
```

mina86 (approved):

PR #179, opened 2024-01-05
solana-ibc: hashing denom for deriving escrow accounts

dhruvja:


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -127,7 +127,7 @@ pub mod solana_ibc {
         ctx: Context<'a, 'a, 'a, 'info, InitEscrow<'info>>,
         port_id: ibc::PortId,
         channel_id_on_b: ibc::ChannelId,
-        base_denom: String,
+        hashed_base_denom: Vec<u8>,
```
mina86, line 130: Why not fixed size hash? Also should this now be called `hashed_denom`?
```suggestion
        hashed_denom: CryptoHash,
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -552,9 +557,9 @@ pub struct SendTransfer<'info> {
     mint_authority: Option<UncheckedAccount<'info>>,
     #[account(mut)]
     token_mint: Option<Box<Account<'info, Mint>>>,
-    // Splitting `base_denom` since each seed can be at most 32 byte long.
+    // Splitting `hashed_base_denom` since each seed can be at most 32 byte long.
     #[account(init_if_needed, payer = sender, seeds = [
-        port_id.as_bytes(), channel_id.as_bytes(), base_denom[..32].as_bytes(), base_denom[32..].as_bytes()
+        port_id.as_bytes(), channel_id.as_bytes(), hashed_base_denom[..32].as_ref(), hashed_base_denom[32..].as_ref()
```
mina86, line 562: Hash is always 32 bytes so no need for splitting
```suggestion
        port_id.as_bytes(), channel_id.as_bytes(), hashed_base_denom.as_bytes()
```

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -87,6 +88,8 @@ fn anchor_test_deliver() -> Result<()> {
     let chain =
         Pubkey::find_program_address(&[crate::CHAIN_SEED], &crate::ID).0;
 
+    let hashed_denom = hash::hash(BASE_DENOM.as_bytes()).to_bytes().to_vec();
```
mina86, line 91: As before, why not CryptoHash?
```suggestion
    let hashed_denom = CryptoHash::digest(BASE_DENOM.as_bytes());
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -39,9 +39,10 @@ fn get_escrow_account(
...
 ) -> Pubkey {
-    let (head, tail) = stdx::split_at::<32>(denom.as_bytes())
+    let denom = hash::hash(denom.as_bytes()).to_bytes();
+    let (head, tail) = stdx::split_at::<32>(&denom)
         .map(|(head, tail)| (&head[..], tail))
-        .unwrap_or((denom.as_bytes(), &b""[..]));
+        .unwrap_or((&denom, &b""[..]));
     let seeds = [port_id.as_bytes(), channel_id.as_bytes(), head, tail];
     let seeds = if tail.is_empty() { &seeds[..3] } else { &seeds[..] };
     Pubkey::find_program_address(seeds, &crate::ID).0
```
mina86, line 48: No need to split since hash is always 32 bytes.
```suggestion
    let denom = CryptoHash::digest(denom.as_bytes());
    let seeds = [port_id.as_bytes(), channel_id.as_bytes(), denom.as_slice()];
    Pubkey::find_program_address(&seeds, &crate::ID).0
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
mina86: Right, but as per Slack discussion, don’t we want to include full denom?

dhruvja: Yes i think we might need full denom. we might have to add trace prefix when we receive the packet since ibc seems to be removing.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -39,12 +39,9 @@ fn get_escrow_account(
...
     denom: &str,
 ) -> Pubkey {
-    let (head, tail) = stdx::split_at::<32>(denom.as_bytes())
-        .map(|(head, tail)| (&head[..], tail))
-        .unwrap_or((denom.as_bytes(), &b""[..]));
-    let seeds = [port_id.as_bytes(), channel_id.as_bytes(), head, tail];
-    let seeds = if tail.is_empty() { &seeds[..3] } else { &seeds[..] };
-    Pubkey::find_program_address(seeds, &crate::ID).0
+    let denom = hash::hash(denom.as_bytes()).to_bytes();
+    let seeds = [port_id.as_bytes(), channel_id.as_bytes(), &denom];
```
mina86, line 43: Let’s do CryptoHash here as well so we consistently use this type
everywhere.

```suggestion
    let denom = lib::hash::CryptoHash::digest(denom.as_bytes());
    let seeds = [port_id.as_bytes(), channel_id.as_bytes(), denom.as_slice()];
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -330,10 +327,8 @@ impl IbcStorage<'_, '_> {
             return Err(TokenTransferError::ParseAccountFailure);
         }
 
-        // Splitting the denom because the trace prefix is not stripped during `send_transfer`.
         let denom = coin.denom.to_string();
-        let denom = denom.rsplit_once('/').unwrap_or((&denom, &denom)).1;
-        let escrow = get_escrow_account(port_id, channel_id, denom);
+        let escrow = get_escrow_account(port_id, channel_id, &denom);
```
mina86, line 331: So for the time being should we be using just the base denom here? `coin.denom.base_denom`?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: yes until we figure out the prefix thing.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: yes makes sense.

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -643,10 +643,10 @@ fn anchor_test_deliver() -> Result<()> {
 
     let denom = format!("{}/{channel_id_on_a}/{send_denom}", port_id.clone());
     let hashed_denom = CryptoHash::digest(send_denom.as_bytes());
-    let hashed_base_denom =
+    let base_denom =
         ibc::apps::transfer::types::BaseDenom::from_str(&denom).unwrap();
```
mina86, line 647: I’m looking into this right now and I’m pretty sure this should be
something along the lines of:

```rust
    let base_denom = mint_keypair.pubkey().to_string();
    let hashed_denom = CryptoHash::digest(base_denom.as_bytes());
    let denom = format!("{port_id}/{channel_id_on_a}/{base_denom}");
    let denom = ibc::apps::transfer::types::PrefixedDenom::from_str(&denom).unwrap();
    let token = ibc::apps::transfer::types::Coin {
        denom,
        amount: TRANSFER_AMOUNT.into(),
    };
```

(notice change from BaseDonem::from_str to PrefixedDenom::from_str).
However, this does break the test.

dhruvja: it makes sense to use `PrefixedDenom` instead of `BaseDenom`. It is good that the test is failing because since we were using `BaseDenom`, it was assuming that the packet is sent on source chain and unescrowing tokens but as per the current denom, it should have burned tokens instead. I have made the changes and it seems to be working.

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -637,16 +637,16 @@ fn anchor_test_deliver() -> Result<()> {
      * Creating Token Mint
      */
 
-    println!("\nSend Transfer");
+    println!("\nSend Transfer On Source Chain");
 
     let send_denom = mint_keypair.pubkey().to_string();
 
-    let denom = format!("{}/{channel_id_on_a}/{send_denom}", port_id.clone());
+    let denom = format!("{}/{channel_id_on_b}/{send_denom}", port_id.clone());
```
mina86, line 644:
```suggestion
    let denom = format!("{port_id}/{channel_id_on_b}/{send_denom}");
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -328,8 +328,8 @@ impl IbcStorage<'_, '_> {
         }
 
         let denom = coin.denom.to_string();
-        let denom = denom.rsplit_once('/').unwrap_or((&denom, &denom)).1;
-        let escrow = get_escrow_account(port_id, channel_id, &denom);
+        let denom = denom.rsplit_once('/').unwrap_or(("", &denom)).1;
+        let escrow = get_escrow_account(port_id, channel_id, denom);
```
mina86: I believe this can now just be
```sugestion
        let escrow = get_escrow_account(port_id, channel_id, coin.denom.base_denom.as_str());
```

mina86 (approved):

.github/workflows/master.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          toolchain: nightly
```
mina86, line 56: We’re already using nightly, see line 27

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: well, the issue still persists because while unescrowing ( when we receive the packet ), we only get the `base_denom`. So we either use `base_denom` or we try to add prefix which i am wondering how since we dont have the source channel id.

mina86: Right, but I mean at this point this is equivalent. We’re taking only
the part after final slash which is base denom. So we can just take
coin.denom.base_denom.  Though I guess it’s worth adding a TODO:

```rust
        // TODO(#180): Should we use full denom including prefix?
        let escrow = get_escrow_account(port_id, channel_id, coin.denom.base_denom.as_str());
```

dhruvja: sure sure, sounds good, will add it

PR #181, opened 2024-01-09
solana-ibc: end to end tests and remove redunadancies

dhruvja:
We were having a `init` methods where we were creating escrow accounts, but those escrow accounts were being created for wrapped tokens which were wrong. Since only native tokens are escrowed, they should be created for native token mint which will happen in the `send_transfer` method. Made the tests end to end where we escrow, mint, burn and then unescrow the same tokens.

### Changes:
Removed `mock_init_escrow` method and removed `escrow_account` from the init methods as well.
Renamed `init_escrow` to `init_mint`, since it would be called for initializing the wrapped token mint.
Shuffled the tests cases so that they are in proper order.

### Doubt
I observed that ibc uses `channel_id_on_b` for deriving escrow accounts when receiving packet and `channel_id_on_a` when sending a packet. So this would mean that whenever we receive a packet on source chain, the `channel_id_on_a` should be of destination ( i think this would be handled by the relayer ).

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/tests.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/tests.rs
@@ -302,7 +302,7 @@ fn anchor_test_deliver() -> Result<()> {
         .instruction(ComputeBudgetInstruction::set_compute_unit_limit(
             1_000_000u32,
         ))
-        .accounts(accounts::InitEscrow {
+        .accounts(accounts::InitMint {
             sender: authority.pubkey(),
             mint_authority: mint_authority_key,
             // escrow_account: escrow_account_key,
@@ -311,7 +311,7 @@ fn anchor_test_deliver() -> Result<()> {
             associated_token_program: anchor_spl::associated_token::ID,
             token_program: anchor_spl::token::ID,
         })
-        .args(instruction::InitEscrow {
+        .args(instruction::InitMint {
             port_id: port_id.clone(),
             channel_id_on_b: channel_id_on_a.clone(),
             hashed_base_denom: hashed_denom.clone(),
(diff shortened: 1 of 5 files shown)


mina86:
> Since only native tokens are escrowed,

Is that true though? Someone may try and send wrapped tokens through our channel, no? They may have wrapped tokens sent through some other bridge for example. Or some other non-SOL token.

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -469,11 +451,12 @@ fn anchor_test_deliver() -> Result<()> {
      */
 
     println!("\nRecving on destination chain");
-    let account_balance_before = sol_rpc_client
-        .get_token_account_balance(&receiver_token_address)
-        .unwrap();
+    let account_balance_before =
+        sol_rpc_client.get_token_account_balance(&receiver_token_address);
+    // .unwrap();
```
mina86, line 456: Why aren’t we unwrapping?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/tests.rs
dhruvja: Because the account is not yet created, it will be created in the following instruction. But we could also just skip it and use 0 instead.

mina86: `.unwrap_or_default()` perhaps? Or whatever appropriate `unwrap_or` would be in this case. Though if we expect this to fail (because account is not there yet), let’s just remove this call?

mina86:
```rust
    let account_balance_before = sol_rpc_client
        .get_token_account_balance(&receiver_token_address)
        .map_or(0, |balance| balance.ui_amount.unwrap());
```

should work.  If not assuming zero sounds good to me.

dhruvja: LGTM, lets use the above.

PR #182, opened 2024-01-10
solana-ibc: add CPI verification with method to get validators.

dhruvja:
For updating the validators stake, `set_stake` will be called from the
restaking program.  To verify if the call is really being made from
it, we pass the instruction key from which we can verify if the call
is really being made from the expected program (restaking program).

Added a method to fetch the current validators from which we can find
their stake and rewards (when we add it).  This method is useful to
get rewards when the user wants the withdraw the funds.

### DIFF
--- a/common/blockchain/src/candidates.rs
+++ b/common/blockchain/src/candidates.rs
@@ -39,7 +39,7 @@ pub struct Candidate<PK> {
     pub pubkey: PK,
 
     /// Candidate’s stake.
-    stake: NonZeroU128,
+    pub stake: NonZeroU128,
 }
 
 /// Error while updating candidate.
--- a/common/blockchain/src/candidates.rs
+++ b/common/blockchain/src/candidates.rs
@@ -34,12 +34,12 @@ pub struct Candidates<PK> {
 #[derive(
     Clone, PartialEq, Eq, borsh::BorshSerialize, borsh::BorshDeserialize,
 )]
-pub struct Candidate<PK> {
+struct Candidate<PK> {
     /// Public key of the candidate.
-    pub pubkey: PK,
+    pubkey: PK,
 
     /// Candidate’s stake.
-    pub stake: NonZeroU128,
+    stake: NonZeroU128,
 }
 
 /// Error while updating candidate.
@@ -90,10 +90,6 @@ impl<PK: crate::PubKey> Candidates<PK> {
         let this = Self { max_validators, candidates, changed, head_stake };
         this.debug_verify_state();
         this
-    }
-
-    pub fn get_candidates(&self) -> Vec<Candidate<PK>> {
-        self.candidates.clone()
     }
 
     /// Sums stake of the first `count` candidates.
--- a/common/blockchain/src/candidates.rs
+++ b/common/blockchain/src/candidates.rs
@@ -34,9 +34,9 @@ pub struct Candidates<PK> {
 #[derive(
     Clone, PartialEq, Eq, borsh::BorshSerialize, borsh::BorshDeserialize,
 )]
-struct Candidate<PK> {
+pub struct Candidate<PK> {
     /// Public key of the candidate.
-    pubkey: PK,
+    pub pubkey: PK,
 
     /// Candidate’s stake.
     stake: NonZeroU128,
@@ -90,6 +90,10 @@ impl<PK: crate::PubKey> Candidates<PK> {
         let this = Self { max_validators, candidates, changed, head_stake };
         this.debug_verify_state();
         this
+    }
+
+    pub fn get_candidates(&self) -> Vec<Candidate<PK>> {
+        self.candidates.clone()
     }
 
     /// Sums stake of the first `count` candidates.
--- a/common/blockchain/src/manager.rs
+++ b/common/blockchain/src/manager.rs
@@ -284,7 +284,7 @@ impl<PK: crate::PubKey> ChainManager<PK> {
         self.candidates.update(&self.config, pubkey, stake)
     }
 
-    pub fn get_validators(&self) -> &[Validator<PK>] {
+    pub fn validators(&self) -> &[Validator<PK>] {
         self.next_epoch.validators()
     }
 }
--- a/common/blockchain/src/manager.rs
+++ b/common/blockchain/src/manager.rs
@@ -1,13 +1,12 @@
 #[cfg(not(feature = "std"))]
 use alloc::collections::BTreeSet as Set;
-use alloc::vec::Vec;
 use core::num::{NonZeroU128, NonZeroU64};
 #[cfg(feature = "std")]
 use std::collections::HashSet as Set;
 
 use lib::hash::CryptoHash;
 
-use crate::candidates::Candidate;
+use crate::Validator;
 pub use crate::candidates::UpdateCandidateError;
 
 #[derive(Clone, Debug, borsh::BorshSerialize, borsh::BorshDeserialize)]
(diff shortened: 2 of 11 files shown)


dhruvja:
Please ignore the change in `programId`. I had to change it since i accidentally ran `cargo clean` which deleted the program id keypair.

solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -150,6 +150,23 @@ impl ChainData {
...
+        ChainNotInitialised,
+    > {
+        let inner = self.get()?;
+        Ok(inner
+            .manager
+            .get_candidates()
+            .iter()
+            .find(|&c| c.pubkey == validator.into())
+            .cloned())
+    }
```
mina86, line 168: Candidate isn’t the same as a validator. If you need to find a validator by a pubkey, you need to look at  ChainManager::next_epoch field.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -111,7 +115,24 @@ pub mod solana_ibc {
     /// TODO(mina86): At the moment we’re operating on pretend tokens and each
     /// validator can set whatever stake they want.  This is purely for testing
     /// and not intended for production use.
+    ///
+    /// Can only be called through CPI from our staking program
+    /// (Program ID: 4EgHMraeMbgQsKyx7sG81ovudTkYN3XcSHpYAJayxCEG)
```
mina86, line 120: I’d consider adding the program id to ChainData so that it doesn’t have to be hard-coded.

common/blockchain/src/candidates.rs
```diff
@@ -92,6 +92,10 @@ impl<PK: crate::PubKey> Candidates<PK> {
         this
     }
 
+    pub fn get_candidates(&self) -> Vec<Candidate<PK>> {
+        self.candidates.clone()
+    }
```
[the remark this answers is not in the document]
dhruvja: got it

common/blockchain/src/manager.rs
```diff
@@ -282,6 +284,10 @@ impl<PK: crate::PubKey> ChainManager<PK> {
     ) -> Result<(), UpdateCandidateError> {
         self.candidates.update(&self.config, pubkey, stake)
     }
+
+    pub fn get_candidates(&self) -> Vec<Candidate<PK>> {
+        self.candidates.get_candidates()
+    }
```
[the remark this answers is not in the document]
dhruvja: yess

solana/solana-ibc/programs/solana-ibc/src/chain.rs
dhruvja: oh i see, makes sense.

common/blockchain/src/manager.rs
```diff
@@ -282,6 +283,10 @@ impl<PK: crate::PubKey> ChainManager<PK> {
     ) -> Result<(), UpdateCandidateError> {
         self.candidates.update(&self.config, pubkey, stake)
     }
+
+    pub fn get_validators(&self) -> &[Validator<PK>] {
```
mina86, line 287: Typically it’s more Rustic to not use ‘get_’ prefix for getters.
```suggestion
    pub fn validators(&self) -> &[Validator<PK>] {
```

solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -172,6 +198,9 @@ struct ChainInner {
 
     /// The guest blockchain manager handling generation of new guest blocks.
     manager: Manager,
+
+    /// Staking Contract program ID. The program which would make CPI calls to set the stake
+    staking_program_id: Box<Pubkey>,
```
mina86, line 203: Any reason to box it? We’re already boxing ChainInner.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -111,10 +118,29 @@ pub mod solana_ibc {
...
```
mina86, line 123:
```suggestion
    /// Can only be called through CPI from our staking program whose
    /// id is stored in chain account.
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -324,6 +350,10 @@ pub struct Chain<'info> {
     trie: UncheckedAccount<'info>,
 
     system_program: Program<'info, System>,
+
+    ///CHECK:
+    // Used for getting the caller program id to verify if the right program is calling the method.
```
mina86, line 355:
```suggestion
    /// CHECK: Used for getting the caller program id to verify if the right
    /// program is calling the method.
```

solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -150,6 +155,27 @@ impl ChainData {
...
+            .cloned())
+    }
+
+    pub fn get_staking_program_id(
+        &self,
+    ) -> Result<Pubkey, ChainNotInitialised> {
+        let chain_inner = self.get()?;
+        Ok(*chain_inner.staking_program_id)
+    }
+
```
mina86, line 178: Firstly, as before, Rustitc way to name this would be to not use
`get_` prefix.

Secondly, alternative would be to have a `check_staking_program`
method instead:

```suggestion
    /// Checks whether given `program_id` matches expected staking program id.
    ///
    /// The staking program id is stored within the chain account.  Various
    /// CPI calls which affect stake and rewards can only be made from that
    /// program.  This method checks whether program id given as argument
    /// matches the one we expect.  If it doesn’t, returns `InvalidCPICall`.
    pub fn check_staking_program(
        &self,
        program_id: &Pubkey
    ) -> Result<(), Error> {
        match program_id == &self.get()?.staking_program_id {
            false => Err(Error::InvalidCPICall),
            true => Ok(()),
        }
    }

```

Dunno if this is necessarily better.  Up to you.

mina86 (approved):

common/blockchain/src/manager.rs
dhruvja: oh i see.

solana/solana-ibc/programs/solana-ibc/src/chain.rs
dhruvja: because we were running out stack on the `deliver` method

solana/solana-ibc/programs/solana-ibc/src/chain.rs
dhruvja: lets go with the above method.

solana/solana-ibc/programs/solana-ibc/src/chain.rs
mina86: But since ChainInner is already on heap, how does putting staking_program_id in a box affect it?

dhruvja: as u said before, when we put the entire `chainInner` in heap, it doesnt allocate everything heap. So maybe it was not allocating `staking_program_id` on heap and hence we ran out of stack since we are on the verge.

mina86: I guess I’m surprised ’cuz it shouldn’t really affect deliver method.

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/chain.rs
dhruvja: It is surprising indeed. There are no accounts which can be boxed as well. But since the stack overflow error is thrown only when i introduced the program id in the `chainInner` struct so i guess it has to be it.

PR #184, opened 2024-01-11
restaking: added restaking contract

dhruvja:


dhruvja:
I am trying to figure on how we could run both rust and ts tests since i was not able to append the ts test command to the existing cargo test command using `;` or `&&`.

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+    pub rent: Sysvar<'info, Rent>,
+}
+
+#[account]
+#[derive(InitSpace)]
+pub struct StakingParams {
+    pub admin: Pubkey,
+    #[max_len(20)]
+    pub whitelisted_tokens: Vec<Pubkey>,
+    pub bounding_timestamp: u64,
```
mina86, line 452: Let’s do i64 so we don’t have to convert between type used by Solana.  Also, let’s suffix it with unit.

```suggestion
    pub bounding_timestamp_sec: i64,
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+
+/// Unused for now
+#[derive(AnchorDeserialize, AnchorSerialize, Clone, Debug)]
+pub enum Service {
+    GuestChain { validator: Pubkey },
+}
+
+#[account]
+pub struct Vault {
+    pub stake_timestamp: u64,
```
mina86, line 464: Ditto.

```suggestion
    pub stake_timestamp_sec: i64,
```

test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+cargo test --lib -- --nocapture --include-ignored ::anchor
+yarn run ts-mocha -p ./tsconfig.json -t 1000000 solana/restaking/tests/**/*.ts
```
mina86:
```suggestion
yarn run ts-mocha -p ./tsconfig.json -t 1000000 solana/restaking/tests/**/*.ts
```

If you want to make it POSIX compliant than:

```sh
#!/bin/sh
cargo test --lib -- --nocapture --include-ignored ::anchor
find solana/restaking/tests/ -name '*.ts' \
     -exec yarn run ts-mocha -p ./tsconfig.json -t 1000000 {} +
```

Either way, I’m not sure if `test.sh` is really the best name for this
file.  It only does Solana tests so `test-solana.sh` seems more
appropriate.

tsconfig.json
```diff
@@ -10,6 +10,8 @@
     "moduleResolution": "node",
     "module": "es2015"
   },
-  "include": ["src/client/**/*"],
+  "resolveJsonModule": true,
+  "esModuleInterop": true,
+  "include": ["/**/*.ts"],
   "exclude": ["node_modules"]
-}
+}
```
mina86, line 18:
```suggestion
}
```

Anchor.toml
```diff
@@ -2,28 +2,41 @@
...
 
-[provider]
-cluster = "Localnet"
-wallet = "~/.config/solana/id.json"
-
 [scripts]
-# test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/**/*.ts"
-test = "cargo test --lib -- --nocapture --include-ignored ::anchor"
+# test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 solana/restaking/tests/**/*.ts"
+test = "sh test.sh"
```
mina86, line 28: Make `test.sh` executable and then:
```suggestion
test = "./test.sh"
```
`sh` is not correct here since `test.sh` is a Bash script. Alternatively, convert `test.sh` to be POSIX compliant.

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+        Ok(())
+    }
+
+    /// We are sending the accounts needed for making CPI call to guest blockchain as [`remaining_accounts`]
+    /// since we were running out of stack memory. Since remaining accounts are not named, they have to be
+    /// sent in the same order as given below
+    /// - SolanaIBCStorage
+    /// - Chain Data
+    /// - trie
+    /// - Guest blockchain program ID
```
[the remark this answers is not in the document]
dhruvja: It offers us validation for accounts and also initializes accounts as we need. But there is a limit and in those cases we have to go with `remaining_accounts`. The issue is only when we `init` the accounts which prolly takes a quite a bit of stack space.

Anchor.toml
dhruvja: makes sense.

solana/restaking/programs/restaking/src/lib.rs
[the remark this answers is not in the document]
dhruvja: I feel its happening only in one method `deposit` due to init but in other methods, we have the same number of accounts and it works well. I feel we have to write a lot of stuff if we drop anchor and tbh not worth it for a single method.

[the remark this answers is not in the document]
dhruvja: like we can always use `remaining_acccounts` when we run out of stack.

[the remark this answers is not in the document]
mina86: Sure, but at this point I feel Account is hindering us more than helping us. It has rigid entry point structure, the way it deals with zerocopy is also quite inflexible.

.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules
 package-lock.json
 target
 test-ledger
+*.so
```
mina86, line 11: Aren’t .so files inside of target directory which is ignored?

test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+cargo test --lib -- --nocapture --include-ignored ::anchor
```
mina86, line 2: Also include `set -eux` as the first instruction. Otherwise, cargo test failing won’t fail the test.

dhruvja: got it

solana/restaking/programs/restaking/src/lib.rs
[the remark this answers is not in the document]
dhruvja: Well i agree with that. But i really dont feel the need to move from anchor YET!

test.sh
dhruvja: yes yes `test-solana.sh` is a better name.

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            ctx.accounts.depositor.to_account_info(),
+            ctx.accounts.depositor.to_account_info(),
+            ctx.accounts.receipt_token_account.to_account_info(),
+            ctx.accounts.token_program.to_account_info(),
+            ctx.accounts.metadata_program.to_account_info(),
+            ctx.accounts.depositor.to_account_info(),
+            ctx.accounts.system_program.to_account_info(),
+            ctx.accounts.rent.to_account_info(),
+            ctx.accounts.nft_metadata.to_account_info(),
+            ctx.accounts.master_edition_account.to_account_info(),
```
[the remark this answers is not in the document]
dhruvja: sure thing.

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            [STAKING_PARAMS_SEED, TEST_SEED, core::slice::from_ref(&bump)];
+        let seeds = seeds.as_ref();
+        let seeds = core::slice::from_ref(&seeds);
+
+        // Transfer the tokens from the platfrom rewards token account to the user token account
+        transfer(
+            ctx.accounts.platform_rewards_token_account.to_account_info(),
+            ctx.accounts.depositor_rewards_token_account.to_account_info(),
+            ctx.accounts.staking_params.to_account_info(),
+            ctx.accounts.token_program.to_account_info(),
```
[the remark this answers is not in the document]
dhruvja: sure

solana/restaking/programs/restaking/src/lib.rs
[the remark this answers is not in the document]
mina86: Let’s create a struct then.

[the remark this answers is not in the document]
dhruvja: sure

Cargo.toml
```diff
@@ -66,6 +68,7 @@ lib = { path = "common/lib" }
 memory = { path = "common/memory" }
 sealable-trie = { path = "common/sealable-trie" }
 solana-trie = { path = "solana/trie" }
+solana-ibc = { path = "solana/solana-ibc/programs/solana-ibc" }
```
mina86, line 71:
```suggestion
solana-ibc = { path = "solana/solana-ibc/programs/solana-ibc" }
solana-trie = { path = "solana/trie" }
```

mina86 (approved):

PR #185, opened 2024-01-11
yarn: update start-server-and-test dependency to bump axios

mina86:
Update start-server-and-test to 2.x which results in a newer axios indirect dependency which addresses vulnerability with the package:

> An issue discovered in Axios 0.8.1 through 1.5.1 inadvertently
> reveals the confidential XSRF-TOKEN stored in cookies by including
> it in the HTTP header X-XSRF-TOKEN for every request made to any
> host allowing attackers to view sensitive information.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "eslint-config-prettier": "^6.15.0",
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.1.2",
-    "start-server-and-test": "^1.11.6",
+    "start-server-and-test": "^2.0",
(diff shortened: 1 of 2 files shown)


mina86:
Are we using yarn for solana-ibc at the moment? Maybe we should just delete trie-example instead?

dhruvja:
> Are we using yarn for solana-ibc at the moment? Maybe we should just delete trie-example instead?

we are not using for `solana-ibc`, but we would be needing it for `restaking` contract

dhruvja (approved):

PR #186, opened 2024-01-12
blockchain: introduce helper impl_proto and any_convert macros

mina86:
Add proto::impl_proto macro which implements conversion from a raw
protocol buffer type to Any type and ibc_state::any_convert macro
which implements conversion between Rust message type and Any type.
For now, use them with ConsensusState.  In the future they will also
be used for ClientState.


dhruvja (approved):
LGTM

PR #188, opened 2024-01-17
restaking: depositing before guest chain is initialized and fix vulnerabilites

dhruvja:
Before the guest chain is initialized ( Bridge is launched ), users can stake their assets without choosing the validator since the validators don't exist yet. The staked tokens get locked in the vault and then when the bridge is launched, the users can set the stake on validator through the following options
- Choose a validator randomly
- Choose a validator from the list of top 10 validators chosen by us
- Choose a particular validator.

Also added `update_guest_chain_initialization` which would make the `is_guest_chain_initialized` as `true`.

The update methods are made seperate to isolate them such that one update doesn't fuck up the others.

Also fixed vulnerabilities from the audit.
1. The `Initialize` instruction can be called multiple times due to the use of `init_if_needed`, enabling anyone to alter staking parameters. So using `Init` instead.
2. The optional `service` parameter in the `deposit` instruction may result in fund lockup as it’s currently checked to be not `None` during `withdrawal`. Users can call `set_service` and set the validator if not set before. 
3. Validation checks for the `remaining_accounts` are missing in both the `deposit` instruction and the `solana_ibc::cpi::set_stake function`. The `guest_chain_program_id` is set when the chain is initialized. The `PDA` seeds along with program ID are verified.
4. In the `Deposit` instruction, the `solana_ibc::cpi::set_stake` function lacks parameters related to the mint of the staked amount. So the `stake_mint` account is sent which expects mint to be of `9` decimals.
5. Some `token::transfer` function calls doesn’t need seeds and signed invocation (for example, in deposit instruction). Introducing empty seeds and then checking `seeds.len() == 0` to switch to unsigned invocation when signed invocation is not necessary. Similarly, the burn_nft function call in the Withdraw instruction doesn’t have to be signed.
6. In the `Withdraw instruction`, ncluding `has_one = rewards_token_mint` for the `staking_params` account, aligning with the approach taken in the `Claim` instruction.

### DIFF
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -8,8 +8,8 @@ use solana_ibc::program::SolanaIbc;
 use solana_ibc::CHAIN_SEED;
 
 pub mod constants;
-mod validation;
 mod token;
+mod validation;
 
 use constants::{
     REWARDS_SEED, STAKING_PARAMS_SEED, TEST_SEED, VAULT_PARAMS_SEED, VAULT_SEED,
@@ -77,8 +77,7 @@ pub mod restaking {
         }
 
         let current_time = Clock::get()?.unix_timestamp;
-        let guest_chain_program_id =
-            staking_params.guest_chain_program_id;
+        let guest_chain_program_id = staking_params.guest_chain_program_id;
 
         vault_params.service = service;
         vault_params.stake_timestamp_sec = current_time;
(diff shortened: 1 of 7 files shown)


solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -64,8 +70,17 @@ pub mod restaking {
             .find(|&&token_mint| token_mint == ctx.accounts.token_mint.key())
             .ok_or(error!(ErrorCodes::TokenNotWhitelisted))?;
 
+        staking_params.total_deposited_amount += amount as u128;
+        if staking_params.total_deposited_amount > staking_params.staking_cap {
```
blasrodri, line 74: >=

blasrodri (changes requested):

solana/restaking/programs/restaking/src/lib.rs
dhruvja: it should be fine if its equal right?

blasrodri (approved):

PR #190, opened 2024-01-24
solana-ibc: signature verification for update client using ed25519 program

dhruvja:
Solana programs have a compute budget limit which doesnt allow us to verify ed25519 signatures. So we have to call the `ed22519` program on solana and pass the `message`, `signature` and `pubkey` in the same transaction in which we call the `send_update_client` method on our contract. 

In the `send_update_client` method, we take signature data and compare it with the instruction data which was performed before and verify. If the verification fails we return the error else we execute the `UpdateClient` message. The signature verification wont be done now, since it has already been done before.

**Note:** the `send_update_client` can only be called to execute `UpdateClient` message.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ prost = { version = "0.12.3", default-features = false }
 prost-build = { version = "0.12.3", default-features = false }
 bytemuck = { version = "1.14", default-features = false }
 derive_more = "0.99.17"
+ed25519-consensus = "1"
 hex-literal = "0.4.1"
 ibc = { version = "0.49.1", default-features = false, features = ["borsh", "serde"] }
 ibc-core-channel-types = { version = "0.49.1", default-features = false }
@@ -90,9 +91,7 @@ anyhow = "1.0.32"
 aes-gcm-siv = { git = "https://github.com/RustCrypto/AEADs", rev = "6105d7a5591aefa646a95d12b5e8d3f55a9214ef" }
 curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dalek", rev = "8274d5cbb6fc3f38cdc742b4798173895cd2a290" }
 
-# eyre has a mutable global variable which is something Solana
-# programs cannot have.  tendermint 0.34 enables eyre unconditionally;
-# version which doesn’t do that hasn’t been released yet so we need to
-# refer to a commit on master.
-tendermint = { git = "https://github.com/informalsystems/tendermint-rs", rev = "37822e540e272d2ca9e763769ad20c581203ff9a" }
-tendermint-proto = { git = "https://github.com/informalsystems/tendermint-rs", rev = "37822e540e272d2ca9e763769ad20c581203ff9a" }
+# Skips signature verification since solana programs dont have
+# compute budget to verify ed25519 signatures.
+tendermint = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
+tendermint-proto = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,3 +94,4 @@ curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dal
 # Skips signature verification since solana programs dont have
 # compute budget to verify ed25519 signatures.
 tendermint = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
+tendermint-proto = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,4 +94,3 @@ curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dal
 # Skips signature verification since solana programs dont have
 # compute budget to verify ed25519 signatures.
 tendermint = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
-tendermint-proto = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
--- a/solana/solana-ibc/programs/solana-ibc/src/ed25519.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/ed25519.rs
@@ -86,8 +86,8 @@ impl Verifier {
     /// Returns error if `ix_sysver` is not `AccountInfo` for the Instruction
     /// sysvar, there was no instruction prior to the current on or the previous
     /// instruction was not a call to the Ed25519 native program.
-    pub fn new(ix_sysvar: &AccountInfo<'_>) -> anchor_lang::Result<Self> {
-        let ix = sysvar::instructions::get_instruction_relative(-1, ix_sysvar)?;
+    pub fn new(ix_sysvar: &AccountInfo<'_>, relative_index: i64) -> anchor_lang::Result<Self> {
+        let ix = sysvar::instructions::get_instruction_relative(relative_index, ix_sysvar)?;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -90,9 +91,7 @@ anyhow = "1.0.32"
...
 
-# eyre has a mutable global variable which is something Solana
-# programs cannot have.  tendermint 0.34 enables eyre unconditionally;
-# version which doesn’t do that hasn’t been released yet so we need to
-# refer to a commit on master.
-tendermint = { git = "https://github.com/informalsystems/tendermint-rs", rev = "37822e540e272d2ca9e763769ad20c581203ff9a" }
-tendermint-proto = { git = "https://github.com/informalsystems/tendermint-rs", rev = "37822e540e272d2ca9e763769ad20c581203ff9a" }
+# Skips signature verification since solana programs dont have
+# compute budget to verify ed25519 signatures.
+tendermint = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
```
blasrodri, line 96: let's move this one to composable, and also let's submit a pr upstream

dhruvja: sounds good.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -169,6 +168,67 @@ pub mod solana_ibc {
...
+                        reason: "Signature Verification Failed".to_string(),
+                    }
+                    .into(),
+                )
+                .into());
+            }
+            index += 1;
+        }
+        if cfg!(not(feature = "mocks")) {
+            match update_client_msg.clone() {
```
blasrodri, line 210: why don't you perform this check before the signature verification?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -169,6 +168,67 @@ pub mod solana_ibc {
...
+                    ibc::ClientMsg::UpdateClient(_) => (),
+                    _ => {
+                        panic!("Invalid instruction, expected MsgUpdateClient")
+                    }
+                },
+                _ => panic!("Invalid instruction, expected MsgUpdateClient"),
+            }
+            let mut store = storage::from_ctx!(ctx);
+            let mut router = store.clone();
+            ::ibc::core::entrypoint::dispatch(
```
blasrodri, line 221: how do you ensure that this function can't be called circumventing the sig verification step?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: Because we need to check if signature verification works.

Cargo.toml
```diff
@@ -90,9 +91,7 @@ anyhow = "1.0.32"
...
+tendermint-proto = { git = "https://github.com/dhruvja/tendermint-rs", branch = "skip-sig-verification" }
```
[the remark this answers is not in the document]
dhruvja: actually we dont need it now. We were using because of `eyre` which was causing issue.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: u mean, we need to add a check in `deliver` method so that it cant be called to update the client right?

Cargo.toml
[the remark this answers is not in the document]
dhruvja: We cannot remove it, since there will be a version mismatch

blasrodri (approved):

solana/solana-ibc/programs/solana-ibc/src/ed25519.rs
```diff
@@ -159,22 +165,22 @@ fn verify_impl(
...
+    pub public_key_offset: u16,
     /// Instruction index to find public key.  We support `u16::MAX` only.
-    public_key_instruction_index: u16,
+    pub public_key_instruction_index: u16,
     /// Offset to start of message data
-    message_data_offset: u16,
-    /// Size of message data.
-    message_data_size: u16,
+    pub message_data_offset: u16,
+    /// Size of message data.
```
mina86, line 179:
```suggestion
    /// Size of message data.
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -162,13 +161,78 @@ pub mod solana_ibc {
...
+        ctx: Context<SendUpdateClient>,
+        update_client_msg: ibc::MsgEnvelope,
+        signatures_data: Vec<SignatureData>,
+    ) -> Result<()> {
+        let length = signatures_data.len() as i64;
+        for (index, signature_data) in signatures_data.into_iter().enumerate() {
+            let verifier = ed25519::Verifier::new(
+                &ctx.accounts.ix_sysvar,
+                index as i64 - length,
+            )?;
```
mina86, line 199: This is wrong. All signatures should be verified with a single call to the ed25519 program. Verifier is capable of handling that. You should initialise Verifier once and than call verify for each signature.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -162,13 +161,78 @@ pub mod solana_ibc {
...
+    /// Since we cannot use ed25519 signature verification due to compute budget limit,
+    /// we have to call ed25519 program and pass the message, pubkey and signatures in
+    /// the same transaction in which the below method would be called. We then verify
+    /// the signatures comparing to the data in the previous instructions. If they
+    /// signature verification is success, we send the `UpdateClient` msg. This method
+    /// can only be called to send `UpdateClient` message. Calling otherwise would
+    /// panic.
+    pub fn send_update_client(
+        ctx: Context<SendUpdateClient>,
+        update_client_msg: ibc::MsgEnvelope,
```
mina86, line 191:
```suggestion
        update_client_msg: ibc::MsgUpdateClient,
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -162,13 +161,78 @@ pub mod solana_ibc {
...
+        signatures_data: Vec<SignatureData>,
```
mina86, line 192: This needs to be read from the update_client_msg, otherwise user can pass different signatures in the message and here.  And if we do that, we actually won’t need a separate send_update_client entry point.

dhruvja: ya true, u have a point. Let me try to do that then.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: oh i see, makes sense. Will do that.

dhruvja:
> I think this is insufficient though. Once the message is dispatched to ibc-rs, it will try to verify the signatures anyway. That part of the code would need to be patched instead.

You would then have to send the `instruction_data` to ibc-rs in that case right?

mina86:
Ideally we’d need to pass `ix_sysvar` to ibc-rs somehow and then do the verification inside ibc-rs.

dhruvja:
ya i guess, maybe modifying `MsgUpdateClient`?

dhruvja:
so that would require change only in the ibc-rs i guess. Do you think we can use `target_os = "solana"` feature for this?

mina86:
It’s more complicated than that.  The problem is that to create a verifier we need to pass it AccountInfo of the instructions system program so replacing the field also affects struct’s constructor.  But the struct is constructed from a protobuf so there’s no obvious way how the verifier can be passed.

dhruvja:
I see

dhruvja:
@mina86 should we close this?

mina86:
I guess we can close this one?

dhruvja:
Yes lets close it.

PR #192, opened 2024-01-25
Restaking audit issues

dhruvja:
Fixed the issues caught during the audit of the restaking contract.
- Changed the service parameter from `Option<Service>` to `Service` so that users choose the service once the guest chain is initialized. There is internal logic which keeps the service `None` if the guest chain is not initialized.
- Checking if `receipt_token_account` has balance in `set_service` method.
- removed signed invocation when calling `set_stake` method and using unsigned invocation instead.
- Added validation for `instruction` account in `deposit` and `set_service` method.

### DIFF
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -459,6 +459,7 @@ pub struct Deposit<'info> {
     pub system_program: Program<'info, System>,
     pub rent: Sysvar<'info, Rent>,
 
+    #[account(address = solana_program::sysvar::instructions::ID)]
     ///CHECK:   
     pub instruction: AccountInfo<'info>,
 
@@ -618,6 +619,7 @@ pub struct SetService<'info> {
     pub stake_mint: Account<'info, Mint>,
 
     ///CHECK:   
+    #[account(address = solana_program::sysvar::instructions::ID)]
     pub instruction: AccountInfo<'info>,
 
     pub system_program: Program<'info, System>,
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -52,7 +52,7 @@ pub mod restaking {
     /// - Guest blockchain program ID
     pub fn deposit<'a, 'info>(
         ctx: Context<'a, 'a, 'a, 'info, Deposit<'info>>,
-        service: Option<Service>,
+        service: Service,
         amount: u64,
     ) -> Result<()> {
         let vault_params = &mut ctx.accounts.vault_params;
@@ -79,7 +79,7 @@ pub mod restaking {
         let guest_chain_program_id = staking_params.guest_chain_program_id;
 
         vault_params.service =
-            if guest_chain_program_id.is_some() { service } else { None };
+            if guest_chain_program_id.is_some() { Some(service) } else { None };
         vault_params.stake_timestamp_sec = current_time;
         vault_params.stake_amount = amount;
         vault_params.stake_mint = ctx.accounts.token_mint.key();
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -98,11 +98,6 @@ pub mod restaking {
                 ctx.remaining_accounts,
                 &guest_chain_program_id.unwrap(),
             )?;
-            let bump = ctx.bumps.staking_params;
-            let seeds =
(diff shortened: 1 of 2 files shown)


solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -330,6 +324,10 @@ pub mod restaking {
         if vault_params.service.is_some() {
             return Err(error!(ErrorCodes::ServiceAlreadySet));
         }
+        let token_account = &ctx.accounts.receipt_token_account;
```
blasrodri, line 327: can you move this validation to the beginning of the function?

dhruvja: sure sure

blasrodri (approved):

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -245,6 +245,15 @@ pub mod restaking {
         Ok(())
     }
 
+    pub fn update_admin(ctx: Context<UpdateStakingParams>, new_admin: Pubkey) -> Result<()> {
```
blasrodri: is this properly guarded so that can only be called by the existing adming?

dhruvja: I am using the existing `UpdateStakingParams` struct which was does checks if it is called by the current admin. Same method which existed when the contract was audited.
https://github.com/ComposableFi/emulated-light-client/blob/de7441e36c58f072724939fce772fe97c632a9f0/solana/restaking/programs/restaking/src/lib.rs#L563

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -431,6 +441,9 @@ pub struct Deposit<'info> {
     #[account(mut, seeds = [STAKING_PARAMS_SEED, TEST_SEED], bump)]
     pub staking_params: Box<Account<'info, StakingParams>>,
 
+    /// Only token mint with 9 decimals can be staked for now since the guest chain expects that.
+    /// If a whitelisted token has 6 decimals, it would just be invalid.
```
mina86, line 445:
```suggestion
    /// Only token mint with 9 decimals can be staked for now since
    /// the guest chain expects that.  If a whitelisted token has 6
    /// decimals, it would just be invalid.
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -324,13 +329,18 @@ pub mod restaking {
...
+            return Err(error!(ErrorCodes::InsufficientReceiptTokenBalance));
+        }
         if staking_params.guest_chain_program_id.is_none() {
             return Err(error!(ErrorCodes::OperationNotAllowed));
         }
         if vault_params.service.is_some() {
             return Err(error!(ErrorCodes::ServiceAlreadySet));
         }
 
+
```
mina86, line 343:
```suggestion
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -455,6 +468,7 @@ pub struct Deposit<'info> {
     pub system_program: Program<'info, System>,
     pub rent: Sysvar<'info, Rent>,
 
+    #[account(address = solana_program::sysvar::instructions::ID)]
     ///CHECK:
```
mina86, line 472: Do we still need this CHECK comment? Same below throughout this PR.

dhruvja: I dont think so. I will remove it.

blasrodri:
Please add the two step authority upgrade that the auditors suggested.

dhruvja:
> Please add the two step authority upgrade that the auditors suggested.

sure thing.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: Well actually seems like, we need to add `/// CHECK:` if its `AccountInfo` or `UncheckedAccount` irrespective whether we have the validation or not. It seems dumb but thats the way it is.

mina86 (approved):

mina86:
What’s the status of this?

dhruvja:
The two step admin upgrade needs to be implemented yet but was parked since i had to check if the upgrade wont break the existing data structure. Looks like it doesnt break and i can push it.

dhruvja:
But we can prolly upgrade the contract later when we launch the guest blockchain.

mina86 (approved):

PR #196, opened 2024-01-31
validator: Implemenation

dhruvja:
Created validator executable that listens to new block events from the guest chain and signs the block and sends it to the guest chain with proper logging.

Things to do:
- ~~CLI to build and run the validator~~
- ~~A config file for `rpc_url`, `keypair_file_path` etc.~~
- ~~modularization of the code.~~


common/blockchain/src/manager.rs
```diff
@@ -289,6 +289,8 @@ impl<PK: crate::PubKey> ChainManager<PK> {
     }
 
     pub fn epoch_height(&self) -> crate::HostHeight { self.epoch_height }
+
+    pub fn genesis(&self) -> CryptoHash { self.genesis.clone() }
```
mina86, line 293:
```suggestion
    pub fn genesis(&self) -> &CryptoHash { &self.genesis }
```

Caller can decide whether they want to clone.

solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -184,6 +184,11 @@ impl ChainData {
         Ok((0, u64::from(current_height)))
     }
 
+    pub fn genesis(&self) -> Result<CryptoHash, ChainNotInitialised> {
+        let inner = self.get()?;
+        Ok(inner.manager.genesis())
```
mina86, line 189:
```suggestion
        Ok(inner.manager.genesis().clone())
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+                            tx.unwrap()
+                        );
+                    }
+                }
+            }
+            Err(err) => {
+                panic!("{}", format!("Disconnected: {err}"));
+            }
+        }
+    }
```
mina86, line 117: Use early returns to limit nesting:

```suggestion
    loop {
        let logs = receiver
            .recv()
            .unwrap_or_else(|err| panic!("Disconnected: {err}"));

        let events = get_events_from_logs(logs.value.logs);
        if events.is_empty() {
            continue;
        }

        // Since only 1 block would be created in a transaction
        let event = &events[0];
        log::info!("Found New Block Event {:?}", event);
        // Fetching the pending block fingerprint
        let fingerprint = blockchain::block::Fingerprint::new(
            &CryptoHash::from_base64(&genesis_hash)
                .expect("Invalid Gensis hash"),
            &event.block_header.0,
        );
        let signature =
            validator.sign_message(fingerprint.as_slice());

        // Send the signature
        let tx = program
            .request()
            .instruction(new_ed25519_instruction_with_signature(&validator.pubkey().to_bytes(), signature.as_ref(), fingerprint.as_slice()))
            .accounts(accounts::ChainWithVerifier {
                sender: validator.pubkey(),
                chain,
                trie,
                ix_sysvar: anchor_lang::solana_program::sysvar::instructions::ID,
                system_program: anchor_lang::solana_program::system_program::ID,
            })
            .args(instruction::SignBlock {
                signature: signature.into(),
            })
            .payer(validator.clone())
            .signer(&*validator)
            .send_with_spinner_and_config(
                RpcSendTransactionConfig {
                    skip_preflight: true,
                    ..RpcSendTransactionConfig::default()
                },
            ).map_err(|e| log::error!("Failed to send the transaction {}", e));
        if tx.is_ok() {
            log::info!(
                "Block signed -> Transaction: {}",
                tx.unwrap()
            );
        }
    }
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    log::info!("Validator running");
+
+    loop {
+        match receiver.recv() {
+            Ok(logs) => {
+                let events = get_events_from_logs(logs.value.logs);
+                if !events.is_empty() {
+                    // Since only 1 block would be created in a transaction
+                    let event = &events[0];
```
mina86, line 72: Probably should check or assert that this is the case.
```suggestion
                    // Since only 1 block would be created in a transaction
                    assert_eq!(1, events.len());
                    let event = &events[0];
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+    let rpc_url = std::env::var("RPC_URL")
+        .unwrap_or_else(|_| "http://127.0.0.1:8899".to_string());
+    let ws_url = std::env::var("WS_URL")
+        .unwrap_or_else(|_| "ws://127.0.0.1:8900".to_string());
+    let program_id: String = std::env::var("PROGRAM_ID").unwrap_or_else(|_| {
+        "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX".to_string()
+    });
+    let genesis_hash = std::env::var("GENESIS_HASH").unwrap_or_else(|_| {
+        "o0OboGSWBU0eJmu3A8mraKscUSOm5LaCKRWLW4IAANw=".to_string()
+    });
```
mina86, line 36: Parse it here into CryptoHash so you don’t have to parse it in the loop later.

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+        .filter_map(|log| {
+            if log.starts_with("Program data: ") {
+                Some(log.strip_prefix("Program data: ").unwrap())
+            } else {
+                None
+            }
+        })
+        .collect();
+    let events: Vec<solana_ibc::events::NewBlock> = serialized_events
+        .iter()
```
mina86, line 138: The collect is not necessary, just continue with iteration.

```suggestion
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+    logs: Vec<String>,
+) -> Vec<solana_ibc::events::NewBlock<'static>> {
+    let serialized_events: Vec<&str> = logs
+        .iter()
+        .filter_map(|log| {
+            if log.starts_with("Program data: ") {
+                Some(log.strip_prefix("Program data: ").unwrap())
+            } else {
+                None
+            }
```
mina86, line 134:
```suggestion
            log.strip_prefix("Program data: ")
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+            match decoded_event {
+                solana_ibc::events::Event::NewBlock(e) => Some(e),
+                _ => {
+                    // println!("This is other event");
+                    None
+                }
+            }
+        })
+        .collect();
+    events
```
mina86, line 154:
```suggestion
        .collect()
```

validator/src/main.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    instruction_data.extend_from_slice(pubkey);
+
+    debug_assert_eq!(instruction_data.len(), signature_offset);
+
+    instruction_data.extend_from_slice(signature);
+
+    debug_assert_eq!(instruction_data.len(), message_data_offset);
+
+    instruction_data.extend_from_slice(message);
```
mina86, line 209: I think this can be done simpler with concat:

```suggestion
    let instruction = [
        &[num_signatures, 0],
        bytes_of(&offsets),
        pubkey,
        signature,
        message
    ].concat()
```

common/blockchain/src/manager.rs
dhruvja: gotcha

validator/src/main.rs
dhruvja: yes, looks good in this way

validator/src/main.rs
dhruvja: well then, the `GENESIS_HASH` from the env should also be in `CryptoHash` right?

dhruvja: or do you mean outside the loop?

mina86: Yes. You can CryptoHash::from_base64 here.

validator/src/command.rs
```diff
@@ -0,0 +1,226 @@
...
+                ws_url: cmd.ws_url,
+                program_id: cmd.program_id,
+                genesis_hash: cmd.genesis_hash,
+                keypair,
+                log_level: cmd
+                    .log_level
+                    .unwrap_or(LevelFilter::Info)
+                    .to_string(),
+            };
+            let toml_in_string = toml::to_string(&config).unwrap();
```
mina86, line 203: to_string_pretty? Though to be honest I have no idea what’s the difference between the two.

validator/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+    DATA_START, PUBKEY_SERIALIZED_SIZE, SIGNATURE_SERIALIZED_SIZE,
+};
+use anchor_client::solana_sdk::{self};
+use anchor_lang::solana_program::instruction::Instruction;
+use base64::Engine;
+use bytemuck::bytes_of;
+use directories::ProjectDirs;
+
+fn project_dirs() -> ProjectDirs {
+    ProjectDirs::from("com", "GuestChain", "Validator")
```
mina86, line 15: Should this be?
```suggestion
    ProjectDirs::from("com", "Composable Finance", "Solana Guest Chain Validator")
```

validator/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+        .unwrap_or_else(|| panic!("Invalid home directory!"))
```
mina86, line 16: `expect`?

validator/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+
+pub fn config_file() -> PathBuf {
+    let proj_dirs = project_dirs();
+    let mut data_file = PathBuf::from(proj_dirs.config_dir());
+    fs::create_dir_all(&data_file).unwrap();
+    data_file.push("config.toml");
+    data_file
+}
+
+pub fn setup_logging(log_level: &str) {
```
mina86, line 27: This should take log::LevelFilter as argument and parsing should be done by caller.

validator/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+    ProjectDirs::from("com", "GuestChain", "Validator")
+        .unwrap_or_else(|| panic!("Invalid home directory!"))
+}
+
+pub fn config_file() -> PathBuf {
+    let proj_dirs = project_dirs();
+    let mut data_file = PathBuf::from(proj_dirs.config_dir());
+    fs::create_dir_all(&data_file).unwrap();
+    data_file.push("config.toml");
+    data_file
```
mina86, line 24: I think it’s cleaner if you don’t explicitly create PathBuf:

```suggestion
    let config_dir = proj_dirs.config_dir();
    fs::create_dir_all(config_dir).unwrap();
    config_dir.join("config.toml")
```

validator/src/command.rs
```diff
@@ -0,0 +1,226 @@
...
+            f,
+            "\nrpc_url: {}\nws_url: {}\nprogram_id: {}\ngenesis_hash: \
+             {}\nvalidator_public_key: {}\nlog_level: {}",
+            self.rpc_url,
+            self.ws_url,
+            self.program_id,
+            self.genesis_hash,
+            keypair.pubkey(),
+            self.log_level
+        )
```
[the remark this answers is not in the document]
dhruvja: well i feel its better to display the public key. Displaying the complete key would be a security risk.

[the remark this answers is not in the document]
mina86: Right, but that’s why I’m asking whether we want it to be shown in Debug implementation (which the current derive does).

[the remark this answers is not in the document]
dhruvja: oh i see, well i dont think so.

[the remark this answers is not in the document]
dhruvja: I mean if we are printing the pubkey only for the user to check if its the right keypair file or not. So i dont think printing the complete keypair would help.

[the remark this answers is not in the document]
dhruvja: so i will write the Debug trait to print pubkey like how i did with format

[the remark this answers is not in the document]
mina86: SG.  You’d probably wanna use [debug_struct](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct).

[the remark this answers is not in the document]
dhruvja: sure sure will check that out.

validator/src/command.rs
dhruvja: well for the `toml` i dont think it makes any difference because it seems to be formatted with just `to_string`.

validator/src/command.rs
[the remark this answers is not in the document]
dhruvja: but why exactly would we need wrapper type?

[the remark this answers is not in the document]
mina86: You could then define Debug and Display on the wrapper such that you’d be able to use Debug derive for this type.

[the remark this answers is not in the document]
dhruvja: oh i see, so instead of implement `Display` and `Debug` for `Config` struct, just implementing it for wrapper type would work. Makes sense now.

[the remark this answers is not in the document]
dhruvja: well since the `Config` struct is used to write to the toml config file, they need to implement `Serde`. We would have to write `Serde` traits for the wrapped type which doesnt seem to worth it to me. wdyt?

validator/src/command.rs
mina86: You can do that with a derive.  I don’t really have strong opinions.  For a moment I wondered whether we should be using base64 to encode the keys but then it looks like Solana code generally encodes keys as array of bytes.

dhruvja: Did u meant this comment for the above conversation? If so, the `Keypair` from solana doesnt implement `Serde` so we may have to write the traits.

mina86: Yes, sorry.  So yeah, do whatever you feel is best.  Using Keypair might have an advantage that we don’t have to convert it into something else later and that it gets validated when deserialising.  We could use `#[serde(serialize_with = "...")]` and `#[serde(deserialize_with = "...")]` to implement the serialisation.  At this point it probably ends up being more code but it would be better typed code.

dhruvja: Well sounds good to me. Will write a wrapper type.

dhruvja:
Displaying keypair wrapper type wont print out the private key. If they really want to print the private key, they can convert to `anchor::Keypair` and print it out.

solana/ed25519/src/lib.rs
```diff
@@ -177,22 +177,22 @@ fn exists_impl(
...
+pub struct SignatureOffsets {
     /// Offset to ed25519 signature of 64 bytes.
-    signature_offset: u16,
+    pub signature_offset: u16,
     /// Instruction index to find signature.  We support `u16::MAX` only.
-    signature_instruction_index: u16,
+    pub signature_instruction_index: u16,
     /// Offset to public key of 32 bytes.
-    public_key_offset: u16,
+    pub public_key_offset: u16,
```
blasrodri, line 186: why do we need an offset?

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -64,7 +66,7 @@ macro_rules! make_message {
 #[test]
 #[ignore = "Requires local validator to run"]
 fn anchor_test_deliver() -> Result<()> {
-    let authority = Rc::new(Keypair::new());
+    let authority = Rc::new(read_keypair_file("../../keypair.json").unwrap());
```
blasrodri, line 69: why not generating a new keypair?

validator/example.config.toml
```diff
@@ -0,0 +1,6 @@
+rpc_url = "http://127.0.0.1:8899"
+ws_url = "ws://127.0.0.1:8900"
+program_id = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
+genesis_hash = "AXO4arKprlSJUQssh8aJxLIWFX5sObiG2Nd2817cfvY="
+keypair = [48,123,8,80,248,0,217,142,124,193,95,24,168,139,214,136,147,210,168,135,26,36,162,89,150,185,99,191,247,135,78,111,12,8,4,81,129,165,153,230,192,225,51,119,216,14,69,225,73,7,204,144,39,213,91,255,136,38,95,131,197,4,101,186]
```
blasrodri, line 5: @dhruvja this key is rubbish, right?

dhruvja: i just picked it from some random github repo

solana/solana-ibc/programs/solana-ibc/src/tests.rs
dhruvja: I am reading the keypair because i wanted the same keypair in tests and validator executable.

solana/ed25519/src/lib.rs
mina86: That’s how Solana defines the struct.  We don’t really have a say here.  This is copied from solana_sdk.

dhruvja: Reference: https://docs.rs/solana-sdk/latest/src/solana_sdk/ed25519_instruction.rs.html#32-83

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+use anchor_client::solana_sdk::signature::{
+    read_keypair_file, Keypair, Signer,
+};
+use clap::{arg, command, Args, Parser, Subcommand};
+use dialoguer::theme::ColorfulTheme;
+use dialoguer::Input;
+use log::LevelFilter;
+use serde::de::Error as SerdeError;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde_bytes::{ByteBuf as SerdeByteBuf, Bytes as SerdeBytes};
```
mina86, line 14: IMO renaming symbols is generally a bad idea.  It makes understanding the code harder since it reduces locality.  Here I would consider using full name at the point where the symbols are used.  They aren’t used that many times.

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+pub struct Config {
+    pub rpc_url: String,
+    pub ws_url: String,
+    pub program_id: String,
+    pub genesis_hash: String,
+    pub keypair: InnerKeypair,
+    pub log_level: String,
+}
+
+pub struct InnerKeypair(Keypair);
```
mina86, line 28:
```suggestion
#[derive(derive_more::From, derive_more::Into)]
pub struct InnerKeypair(Keypair);
```

and then you don’t have to implement From and Into manually.

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+            Ok(Values::No)
+        } else {
+            Err(format!("Can not parse {}", s))
+        }
+    }
+}
+
+impl Display for Values {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
mina86, line 165:
```suggestion
        core::fmt::Debug::fmt(self, f)
```

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+                 command first.",
+            );
+            let default_config: Config = toml::from_str(&config_data).unwrap();
+            let keypair = if cmd.keypair_path.is_some() {
+                let keypair = read_keypair_file(&cmd.keypair_path.unwrap())
+                    .expect("Unable to read keypair file");
+                keypair.into()
+            } else {
+                default_config.keypair
+            };
```
mina86, line 185:
```suggestion
            let keypair = if let Some(path) = cmd.keypair_path {
                read_keypair_file(&keypair_path)
                    .expect("Unable to read keypair file")
                    .into()
            } else {
                default_config.keypair
            };
```

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+                ws_url: cmd.ws_url.unwrap_or(default_config.ws_url),
+                program_id: cmd.program_id.unwrap_or(default_config.program_id),
+                genesis_hash: cmd
+                    .genesis_hash
+                    .unwrap_or(default_config.genesis_hash),
+                keypair,
+                log_level: cmd
+                    .log_level
+                    .unwrap_or(LevelFilter::Info)
+                    .to_string(),
```
mina86, line 197: Why are we conventing to String? Just keep this as LevelFilter?

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+            };
+            setup_logging(LevelFilter::from_str(&config.log_level).unwrap());
+            Some(config)
+        }
+        Commands::Init(cmd) => {
+            let config_file = config_file();
+            let file = fs::metadata(config_file.clone());
```
mina86, line 204:
```suggestion
            let file = fs::metadata(config_file.as_path());
```
Just taking reference should also work.  No need to clone.

validator/src/command.rs
```diff
@@ -0,0 +1,254 @@
...
+            setup_logging(cmd.log_level.unwrap_or(LevelFilter::Info));
+            if file.is_ok() {
```
mina86, line 206: Actually it looks like all you care is checking whether the file exists.  In that case:
```suggestion
            if config_file.exists() {
```
and fs::metadata call isn’t necessary.

validator/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+
+
+fn main() {
+    let config = command::parse_config();
+
+    if config.is_none() {
+        return;
+    }
+
+    let config = config.unwrap();
```
mina86, line 28:
```suggestion
    let config = match command::parse_config() {
        None => return,
        Some(config) => config,
    };
```

validator/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    log::info!("Validator running");
+
+    let genesis_hash = &CryptoHash::from_base64(&config.genesis_hash)
+        .expect("Invalid Genesis Hash");
+
+    loop {
+        let logs = receiver
+            .recv()
+            .unwrap_or_else(|err| panic!("{}", format!("Disconnected: {err}")));
```
mina86, line 64:
```suggestion
            .unwrap_or_else(|err| panic!("Disconnected: {err}"));
```

validator/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+            .payer(validator.clone())
+            .signer(&*validator)
+            .send_with_spinner_and_config(RpcSendTransactionConfig {
+                skip_preflight: true,
+                ..RpcSendTransactionConfig::default()
+            })
+            .map_err(|e| log::error!("Failed to send the transaction {}", e));
+        if tx.is_ok() {
+            log::info!("Block signed -> Transaction: {}", tx.unwrap());
+        }
```
mina86, line 107:
```suggestion
            });
        match tx {
            Ok(tx) => log::info!("Block signed -> Transaction: {tx}"),
            Err(err) => log::error!("Failed to send the transaction {err}"),
        }
```

validator/src/utils.rs
```diff
@@ -0,0 +1,94 @@
+use std::fs;
+use std::path::PathBuf;
+
+use anchor_client::solana_sdk::ed25519_instruction::{
+    DATA_START, PUBKEY_SERIALIZED_SIZE, SIGNATURE_SERIALIZED_SIZE,
+};
+use anchor_client::solana_sdk::{self};
```
mina86, line 7:
```suggestion
use anchor_client::solana_sdk;
```

validator/src/command.rs
dhruvja: right, makes sense.

validator/src/command.rs
dhruvja: nice...

validator/src/command.rs
dhruvja: The LevelFilter doesnt implement Serde which is needed to write to config file.

validator/example.config.toml
blasrodri: lol

blasrodri:
LGTM

PR #198, opened 2024-02-01
solana-ibc: added deliver with chunks method to deserialize data from account

dhruvja:
Added `deliver_with_chunks` method where the `chunks` account is passed where the data inside the account is deserialized and dispatched to ibc. It sends the data to `write-account` program in chunks and then the account is passed to the `deliver-with-chunks` method in `solana-ibc`. 

The only thing i am wondering is that we need to close the account as soon as the `deliver-with-chunks` method is executed. We cannot close the account in `solana-ibc` since the account is owned by `write-account` program. And AFAIK, the account can only be closed by `write-account` program. Well we can close or we can just delete the data. Maybe we have another instruction with key as `2` which clears all the data in the account.

wdyt @mina86 ?


mina86:
> We cannot close the account in solana-ibc since the account is owned by write-account program. And AFAIK, the account can only be closed by write-account program. 

The account isn't owned by write-account.  The account is owned by the caller.  They can create the account, write data to it, call solana-ibc with it and then delete it.

mina86:
By the way, this is one example where Anchor is counterproductive for us.  Without using Anchor, this would be five lines of code and calling with chunks account would be supported for all of the entry point.

dhruvja:
> > We cannot close the account in solana-ibc since the account is owned by write-account program. And AFAIK, the account can only be closed by write-account program.
> 
> The account isn't owned by write-account. The account is owned by the caller. They can create the account, write data to it, call solana-ibc with it and then delete it.

I dont think so. To modify the account state, the program has to own the account, else the account becomes read only. If you see, when i create the account, the owner is the `write-account` program.
https://github.com/ComposableFi/emulated-light-client/blob/4641cfb9f46fe7a3182cbb44a0dd137a672ef5d2/solana/solana-ibc/programs/solana-ibc/src/tests.rs#L167

mina86:
> To modify the account state, the program has to own the account

Oh for fuck’s sake. Solana is literal garbage. You’re right, ‘if an account is not owned by a program, the program is only permitted to read its data and credit the account.’ So there’s no way to pass the account as writeable when making the call. For some reason I thought that could be done.

So this breaks write-account to be honest because different callers can overwrite each other data which is something we definitely don’t want. We need to change it so that it uses PDA with `seeds = [ caller.pubkey(), arbitrary_caller_provided_key ]` and in addition add option to free the account as per what you’ve suggested.

dhruvja:
I feel we just have to call the `write-account` with instruction to clear the data and then proceed to add chunks before every transaction. This would be more than enough and i dont see the need to use PDA because there we wont be needing 2 accounts at the same time. 

Well actually, coming to think of it we need to use PDA. Because if the account is already initialized with other program we cannot use it. 

But the seeds should be `seeds = [ caller.pubkey(), SOME_CONSTANT ]`, because we can clear the data and use the same account multiple times instead of creating multiple accounts.

mina86:
> I feel we just have to call the write-account with instruction to clear the data and then proceed to add chunks before every transaction.

Unfortunately that’s not enough. A caller is unable to write the data in a single transaction and by doing it in multiple transaction this opens possibility for another caller to modify already written data. So we need to have some kind of ownership mechanism where each caller has its own accounts they can write to.

> But the seeds should be `seeds = [ caller.pubkey(), SOME_CONSTANT ]`, because we can clear the data and use the same account multiple times instead of creating multiple accounts.

Adding a caller-controlled seed adds more flexibility to the caller. It may very well be that caller will always use an empty key but nonetheless IMO there should be part of the seed controlled by the caller to provide more options.

dhruvja:
> IMO there should be part of the seed controlled by the caller to provide more options

What are you implying here?

mina86:
I mean, I’m implying what I’ve written. It’s more flexible for the caller. Maybe caller, for whatever reason, uses two different accounts for two different purposes and don’t want to worry if they ever clash? Maybe caller wants to make two calls to solana-ibc in a single transaction and thus needs to write data to two accounts.

blasrodri:
What's the status of this PR?

dhruvja:
Waiting for review from @mina86 . Added PDA with a new instruction which initializes it.

solana-test.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 set -eux
+solana config set --url http://127.0.0.1:8899
+cd solana/write-account
+cargo build-bpf
+cd .. && cd ..
```
mina86, line 6:
```suggestion
cd ../..
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -169,6 +168,23 @@ pub mod solana_ibc {
...
+    /// Forms the message from the data stored in the chunks account
+    ///
+    /// Since the instructions can only be of maximum 1232 bytes, for large messages,
+    /// we need to split them into chunks and store them in accounts ( who have upto 10Mb)  
+    /// of storage. We can then deserialize the message and send it to ibc-rs.
+    pub fn deliver_with_chunks<'a, 'info>(
+        ctx: Context<'a, 'a, 'a, 'info, DeliverWithChunks<'info>>,
+    ) -> Result<()> {
+        let chunks = ctx.accounts.chunks.try_borrow_data().unwrap();
+        let message = ibc::MsgEnvelope::try_from_slice(&chunks).unwrap();
```
mina86, line 180: I suggest we require chunks account to be length-prefixed, so
something like:

```rust
    let chunks = std::split_at::<4>(chunks).and_then(|(len, chunks)| {
        chunks.get(..u32::from_be_bytes(len) as usize)
    }).ok_or(...)?;
    let message = ...;
```

This will make sure we don’t read some garbage data left within the
account if it has been reused.

solana/write-account/src/lib.rs
```diff
@@ -11,28 +16,47 @@ solana_program::entrypoint!(process_instruction);
 /// The first byte of the `instruction` determines operation to perform.  Format
 /// of the instruction and required accounts depend on that.
 ///
+/// # Create PDA Account
+///
+/// Instruction with discriminant zero is Create and its format is as follows:
+/// ```ignore,text
```
mina86, line 22:
```suggestion
///
/// ```ignore,text
```

solana/write-account/src/lib.rs
```diff
@@ -11,28 +16,47 @@ solana_program::entrypoint!(process_instruction);
...
+/// +-----+---------------+------------+
+/// | 0u8 | data_len: u32 | data: [u8] |
```
mina86, line 24: Should this read seed?

```suggestion
/// | 0u8 | account_size: u32 | seed: [u8] |
```

solana/write-account/src/lib.rs
```diff
@@ -11,28 +16,47 @@ solana_program::entrypoint!(process_instruction);
...
+/// +-----+---------------+------------+
+/// ```
+///
+/// It create a Program Derived Address with the signer key and a constant seed
+/// provided by the user.
+/// Requires 3 accounts in the following order
+/// - Write Account PDA: PDA with seeds as [payer_pubkey, seed].
+/// - Payer account: Should be a signer.
+/// - System program: Account used to create PDA. Should be `11111111111111111111111111111111`
```
mina86, line 33:
```suggestion
/// It creates a Program Derived Address (PDA) with the signer key and
/// provided seed.
///
/// Requires 3 accounts in the following order
/// - Write Account PDA: PDA with seeds as [payer_pubkey, seed].
/// - Payer account: Should be a signer.
/// - System program: Account used to create PDA.
///   Should be `11111111111111111111111111111111`
```

solana/write-account/src/lib.rs
```diff
@@ -11,28 +16,47 @@ solana_program::entrypoint!(process_instruction);
 /// The first byte of the `instruction` determines operation to perform.  Format
 /// of the instruction and required accounts depend on that.
 ///
+/// # Create PDA Account
```
mina86, line 19: Do we need a separate create command? I was imagining that write would create the account if it doesn’t exist. Similarly, write would resize it if it’s too small. I think it would be a simpler interface and require fewer smart contract calls.

solana/write-account/src/lib.rs
```diff
@@ -53,39 +77,90 @@ solana_program::entrypoint!(process_instruction);
...
+    let accounts_iter = &mut accounts.iter();
+    let write_account = next_account_info(accounts_iter)?;
+    let payer = next_account_info(accounts_iter)?;
+    let system_program = next_account_info(accounts_iter)?;
+
+    let account_span = data
+        .unshift_n::<4>()
+        .ok_or(ProgramError::InvalidInstructionData)
+        .and_then(usize_from_bytes)?;
+    let lamports_required = (Rent::get()?).minimum_balance(account_span);
```
mina86, line 107:
```suggestion
    let lamports_required = Rent::get()?.minimum_balance(account_span);
```

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -151,14 +163,85 @@ fn anchor_test_deliver() -> Result<()> {
         ibc::ClientMsg::CreateClient,
         ibc::MsgEnvelope::Client,
     );
+
+    let serialized_msg = message.try_to_vec().unwrap();
```
mina86, line 167: With the chunks account being length-prefixed this would be:

```suggestion
    let serialized_msg = message.try_to_vec().unwrap();
    let serialized_msg = [
        &(serialized_msg.len() as u32).to_le_bytes()[..],
        serialized_msg.as_slice(),
    ].concat();
```

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -151,14 +163,85 @@ fn anchor_test_deliver() -> Result<()> {
...
+    let serialized_msg_len = serialized_msg.len() as u32;
+
+    println!("\nCreating Account to store message chunks");
+
+    let blockhash = sol_rpc_client.get_latest_blockhash().unwrap();
+    let mut discriminant = vec![0];
+    let data_size_in_bytes = serialized_msg_len.to_le_bytes().to_vec();
+    discriminant.extend(data_size_in_bytes);
+    discriminant.extend(WRITE_ACCOUNT_SEED.to_vec());
```
mina86, line 176: You don’t need to convert everything to Vec.  First of all Vec can be
extended from a slice but more importantly you can use concat:

```suggestion
    let data_size_in_bytes = .to_vec();
    let instruction = [
        &[0][..],
        &serialized_msg_len.to_le_bytes()[..],
        &WRITE_ACCOUNT_SEED[..],
    ].concat();
```

solana/write-account/src/lib.rs
```diff
@@ -124,6 +199,13 @@ fn handle_copy_null(dst: &mut [u8], src: &[u8]) -> Result {
         .ok_or(ProgramError::AccountDataTooSmall)
 }
 
+#[derive(BorshDeserialize, BorshSerialize, Debug)]
+pub struct State {
+    pub seed: Vec<u8>,
+    pub bump: u8,
+    pub data: Vec<u8>,
+}
```
mina86, line 207: If we’re using borsh for this we might just as well use borsh for the entire instruction.

solana/write-account/src/lib.rs
dhruvja: well the thing is, we cannot create the account and use it in the same method. So the account has to be created prior to calling the `write` method.

solana/write-account/src/lib.rs
dhruvja: Sorry my bad, it reads seed only.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: sure sure, makes sense.

solana/write-account/src/lib.rs
mina86: Interesting.  How does Anchor do `init_if_needed`? I was under the impression that if the account doesn’t exist, it can be created and then operated on.

dhruvja: even i thought the same, but looks like it creates an instruction and executes it before we call the method.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -179,6 +178,28 @@ pub mod solana_ibc {
...
+        let chunks = ctx.accounts.chunks.try_borrow_data().unwrap();
+        let chunks = stdx::split_at::<4, u8>(&chunks)
+            .and_then(|(len, chunks)| {
+                chunks.get(..u32::from_le_bytes(*len) as usize)
+            })
+            .ok_or(error::Error::InvalidChunksData)?;
+        let message = ibc::MsgEnvelope::try_from_slice(chunks).unwrap();
+        let mut store = storage::from_ctx!(ctx, with accounts);
+        let mut router = store.clone();
+        ::ibc::core::entrypoint::dispatch(&mut store, &mut router, message)
```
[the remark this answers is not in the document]
dhruvja: well i feel, rather than converting we can just use copy and the verifier initialization since the converting would be a rather complication.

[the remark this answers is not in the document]
mina86: Yes, by convert I mean just copy the fields over. DeliveryWithChunks has just one more field than Deliver hence why it’s not just straight copy.

solana/write-account/src/lib.rs
```diff
@@ -53,39 +82,86 @@ solana_program::entrypoint!(process_instruction);
...
+        &system_instruction::create_account(
+            payer.key,
+            write_account.key,
+            lamports_required,
+            account_span as u64,
+            program_id,
+        ),
+        &[payer.clone(), write_account.clone(), system_program.clone()],
+        &[&[payer.key.as_ref(), data, &[bump]]],
+    )?;
```
mina86, line 126: Shouldn’t this be [create_account_with_seed](https://docs.rs/solana-program/latest/solana_program/system_instruction/fn.create_account_with_seed.html) instead?

dhruvja: `create_account_with_seed` seems to be different than creating PDA. But i was thinking, we just need an account with seed and we can create it on client side as well. So maybe we can remove this instruction as well if we use those.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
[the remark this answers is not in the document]
dhruvja: But it would require to clone all the accounts, do u think its better to just add the same `deliver` logic to `deliver_with_chunks` method?

solana/write-account/src/lib.rs
```diff
@@ -53,39 +82,86 @@ solana_program::entrypoint!(process_instruction);
...
+
+    let state = State::try_from_slice(data)
+        .map_err(|_| ProgramError::InvalidInstructionData)?;
+
+    let (pubkey, _bump) = Pubkey::find_program_address(
+        &[payer.key.as_ref(), &state.seed],
+        program_id,
+    );
+
+    if write_account.key != &pubkey {
```
mina86, line 152: Shouldn’t we also compare bump we’ve gotten here with bump user provided in the instruction?

solana/write-account/src/lib.rs
```diff
@@ -53,39 +82,86 @@ solana_program::entrypoint!(process_instruction);
...
+
+    invoke_signed(
+        &system_instruction::create_account(
+            payer.key,
+            write_account.key,
+            lamports_required,
+            account_span as u64,
+            program_id,
+        ),
+        &[payer.clone(), write_account.clone(), system_program.clone()],
```
mina86, line 124: Silly question, documentation has example where `&system_program::ID` is passed as owner (we’re passing `program_id` here). I’m bemused why that would be the case. Do you have an inkling for the reasons?

Also, as far as I understand, we only pass two accounts: payer and write_account.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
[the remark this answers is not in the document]
mina86: I don’t see why we need to clone. We’re just moving stuff from one account to the other. We’re discarding DeliveryWithChunks struct so there shouldn’t be need for clone. Or am I missing something?

solana/write-account/src/lib.rs
blasrodri: Yes, Michal is right
https://medium.com/coinmonks/security-of-solana-smart-contracts-why-you-should-always-validate-pda-bump-seeds-d068095fc772

solana/solana-ibc/programs/solana-ibc/src/lib.rs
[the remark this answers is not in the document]
dhruvja: i dont think we can move since the accounts are under mutable reference. We can only clone perhaps.

solana/write-account/src/lib.rs
dhruvja: i think we are using `program_id` as owner so that we can modify the data but `system_program::ID` would be the owner if its a normal account where u just want to transfer funds. Reference: 

[Only an account owned by system program can transfer via system program.](https://solanacookbook.com/references/accounts.html#transfer)

solana/solana-ibc/programs/solana-ibc/src/lib.rs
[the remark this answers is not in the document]
mina86: Ok, let me take a look. As per chat, I’m now looking into #213 and might end up addressing this PR as well.

solana/write-account/src/lib.rs
dhruvja: ya right right, makes sense.

mina86:
I believe we can close this one.

dhruvja:
Yes lets do that.

PR #200, opened 2024-02-02
solana-allocator: use strict pointer provenance

mina86:
Use pointer arithmetic with strict provenance to make Miri happy and
better test the allocator.  Strict provenance pointer methods are
unstable so copy what we need.

### DIFF
--- a/solana/allocator/src/lib.rs
+++ b/solana/allocator/src/lib.rs
@@ -126,7 +126,7 @@ unsafe impl GlobalAlloc for BumpAllocator {
             // On first call, end_pos is null.  Start allocating past the
             // end_pos.
             ptr = ptr::with_addr(
-                self.ptr.as_ptr(),
+                self.heap_range().start,
                 ptr::end_addr_of_val(end_pos),
             );
         };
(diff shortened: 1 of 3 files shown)


solana/allocator/src/lib.rs
```diff
@@ -21,9 +25,9 @@ use core::cell::Cell;
 /// a segfault once out of available heap memory.
 pub struct BumpAllocator {
     #[cfg(test)]
-    start: core::ptr::NonNull<Cell<usize>>,
+    ptr: core::ptr::NonNull<u8>,
```
dhruvja, line 28: why are these attributes only allowed only in tests?

dhruvja:
Looks like solana VM does support `pointer_byte_offsets` feature.

solana/allocator/src/lib.rs
mina86: In production we’re using Solana provided heap.  `ptr` is only for tests where we allocate chunk of memory and use that.

dhruvja (approved):

PR #204, opened 2024-02-06
solana-ed25519: move ED25519 Solana verification code into a new crate

mina86:
This is mostly because we are going to be patching ibc-rs to support
Solana signature verefication and to avoid circular dependencies we
need the ED25519 code in a separate crate that solana-ibc and ibc-rs
can both depend on.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -24,7 +24,12 @@ jobs:
         id: install-rust
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly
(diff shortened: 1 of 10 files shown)


blasrodri (approved):

mina86:
https://github.com/tkaitchuck/aHash/issues/200 is the clippy failure

mina86:
*sighs* ahash 0.8.6 doesn’t compile on nightly and ahash 0.8.7 doesn’t compile with Anchor’s version of rustc.

PR #206, opened 2024-02-07
solana-ibc: use Solana ed25519 verifier for Tendermint client state

mina86:
Take advantage of the support for global state in the allocator and
use it for the verifier state.  We’re using that state in patched
Tendermint client implementation to verify signatures there.

At the moment, because of the limitations of Solana we can verify less
than 10 signatures, but this change at least demonstrates the approach
we can take to have Tendermint client state updates implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2048,8 +2048,7 @@ dependencies = [
 [[package]]
 name = "ibc"
 version = "0.50.0"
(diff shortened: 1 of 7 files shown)


dhruvja:
I dont think its the account thats causing the problem. Its the `set_verifier` method because when i commented that method, it worked fine.

blasrodri (approved):

PR #213, opened 2024-02-13
write-account: add support for creating and resizing the Write account

mina86:
Since the Write account needs to be a PDA, add support for creating
and resizing it as necessary.  We’re potentially missing Truncate and
Free instructions but we can live without them for the time being.

Furthermore, remove the Copy instruction.  We don’t use it currently
and having support for it while details of creation and resizing of
the account aren’t yet established create unnecessary maintenance
burden.

### DIFF
--- a/solana/write-account/src/lib.rs
+++ b/solana/write-account/src/lib.rs
@@ -26,6 +26,7 @@ solana_program::entrypoint!(process_instruction);
 ///     discriminant: u8,  // always 0u8,
 ///     seed_len: u8,
 ///     seed: [u8; seed_len],
+///     bump: u8,
 ///     offset: u32,
 ///     data: [u8],
 /// }
@@ -49,6 +50,8 @@ solana_program::entrypoint!(process_instruction);
 ///
 /// Note: `data` may be empty in which case the instruction will just create or
 /// resize the Write account.
+// TODO(mina86): Add option to truncate the account and to free the account
+// returning lamports to payer.
 fn process_instruction(
     program_id: &Pubkey,
     accounts: &[AccountInfo],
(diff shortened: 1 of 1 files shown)


solana/write-account/src/lib.rs
```diff
@@ -9,145 +12,227 @@ solana_program::entrypoint!(process_instruction);
...
-    match algo {
-        0 => handle_copy_null(dst, src),
-        _ => Err(ProgramError::InvalidInstructionData),
+    let lamports = write_account.lamports();
+    let get_required_lamports =
+        || Rent::get().map(|rent| rent.minimum_balance(size));
+
+    if lamports == 0 {
+        // If the account has zero lamports it needs to be created first.
+        let _ = system.ok_or(ProgramError::NotEnoughAccountKeys)?;
```
dhruvja, line 160:
```suggestion
        let system_program = system.ok_or(ProgramError::NotEnoughAccountKeys)?;
```

solana/write-account/src/lib.rs
```diff
@@ -9,145 +12,227 @@ solana_program::entrypoint!(process_instruction);
...
+        let instruction = solana_program::system_instruction::create_account(
+            payer.key,
+            write_account.key,
+            get_required_lamports()?,
+            size as u64,
+            program_id,
+        );
+        solana_program::program::invoke_signed(
+            &instruction,
+            &[payer.clone(), write_account.clone()],
```
dhruvja, line 170: I feel creating account requires system program.
```suggestion
            &[payer.clone(), write_account.clone(), system_program.clone()],
```

mina86: Solana is weird. The system program account must be there passed to the smart contract invocation, but it’s not needed to in the invocation. See [example in documentation](https://docs.rs/solana-program/latest/solana_program/system_instruction/fn.create_account.html#example-on-chain-program). I’m also double-checking in #215.

dhruvja: This is strange.

mina86: Anchor Test in #215 passed so this PR is correct. System program doesn’t need to be passed in the invocation.

dhruvja (approved):

PR #215, opened 2024-02-13
solana-ibc: support reading instruction data from account

mina86:
Introduce a mechanism by which clients of Solana IBC can store
instruction data inside of an account and then instruct Solana IBC to
read instruction data from said account rather than the Solana
transaction.

This is necessary because Solana limits transactions to roughly 1.2 kB
which includes full transaction including list of accounts, all the
instructions etc.  The end result is that some of the Solana IBC
instructions simply don’t fit.

Co-authored-by: Dhruv D Jain <dhruv@iamsizzling.com>

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/ix_data_account.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/ix_data_account.rs
@@ -1,0 +1,128 @@
+//! Support for calling Solana IBC smart contract with instruction data read
+//! from an account.
+//!
+//! Solana limits transaction size to at most 1232 bytes.  This includes all
+//! accounts participating in the transaction as well as all the instruction
+//! data.  Unfortunately, with IBC we may need to encode instructions which
+//! don’t fit in that limit.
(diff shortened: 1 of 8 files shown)


solana/solana-ibc/programs/solana-ibc/src/ix_data_account.rs
```diff
@@ -0,0 +1,168 @@
+//! Support for calling Solana IBC smart contract with instruction data read
+//! from an account.
+//!
+//! Solana limits transaction size to at most 1232 bytes.  This includes all
+//! accounts participating in the transaction as well as all the instruction
+//! data.  Unfortunately, with IBC we may need to encode instructions which
+//! don’t fit in that limit.
+//!
+//! To address this, Solana IBc smart contract supports reading instruction data
```
dhruvja, line 9:
```suggestion
//! To address this, Solana IBC smart contract supports reading instruction data
```

dhruvja (approved):

PR #216, opened 2024-02-14
solana-ibc: Delegation, setting appropriate stake amount and fetching candidate during withdrawal

dhruvja:
When the restaking program would call the `set_stake` method in solana ibc, we were setting the stake for the caller address which wouldnt be the case if the stake is being delegated. So supplying the `set stake` method with validator key indicating the validator to which the stake is getting delegated to. 

When setting stake, add the existing stake to the new stake and call the `set_stake` method in `solana_ibc` during deposit. During withdrawal, perform vice versa.

During withdrawal, fetch the stake from the candidate instead of validator since the candidate may or may not be part of validator in the next epoch based on the stake.


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -139,7 +143,13 @@ pub mod solana_ibc {
         chain.check_staking_program(&current_ixn.program_id)?;
         let provable = storage::get_provable_from(&ctx.accounts.trie)?;
         chain.maybe_generate_block(&provable)?;
-        chain.set_stake((*ctx.accounts.sender.key).into(), amount)
+        // Setting the stake to 0 removes the validator from the next
+        // epoch so the validator key passed in argument should be the
+        // signer.
```
mina86, line 148: I don’t understand this. Setting stake to 1 effectively removes validator from the set as well. Why is zero somehow special? Any change to stake should have the same kind of authorisation.

dhruvja: so if the stake is below the minimum stake, validator is gonna be removed?

dhruvja: i mean for delegation, it would be other user who would call right. Since we are calling from restaking program, as long as we make sure the logic is right, we can remove the above logic.

mina86: I mean, even if the stake is above minimum, if there’s a lot of candidates, if one validator’s stake is set too low, they will be removed from the set of validators because there will be someone else with more stake.

My understanding was that it’s always restaking program which calls set_stake.  Restaking is holding the funds and tracking who delegates them.

dhruvja: Yes, but previously we were just setting the stake on the account who is calling the set_stake, so i think we need an argument to pass the validator to which it is getting delegated.

dhruvja: But what i am missing is updating the new stake with the previous one.

mina86: > Yes, but previously we were just setting the stake on the account who is calling the set_stake, so i think we need an argument to pass the validator to which it is getting delegated.

Yes, I get that. But what I don’t get is why there’s a special case for `amount == 0`. It seems to me that all checking needs to be done in restaking program since it has all the state and accepts deposits.

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -94,18 +94,37 @@ pub mod restaking {
 
         // Call Guest chain program to update the stake if the chain is initialized
         if guest_chain_program_id.is_some() {
+            let service =
+                service.as_ref().ok_or(error!(ErrorCodes::MissingService))?;
```
mina86, line 98: Do we want lazy evaluation here?
```suggestion
                service.as_ref().ok_or_else(|| error!(ErrorCodes::MissingService))?;
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -94,18 +94,37 @@ pub mod restaking {
...
+            let validator_key = match service {
+                Service::GuestChain { validator } => validator,
+            };
+            let borrowed_chain_data =
+                ctx.remaining_accounts[0].data.try_borrow_mut().unwrap();
```
mina86, line 103: Why borrow mut if you’re only taking shared reference later on?
```suggestion
                ctx.remaining_accounts[0].data.try_borrow().unwrap();
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -94,18 +94,37 @@ pub mod restaking {
...
+                solana_ibc::chain::ChainData::try_deserialize(&mut chain_data)
+                    .unwrap();
+            let validator = chain
+                .validator(*validator_key)
+                .map_err(|_| ErrorCodes::OperationNotAllowed)?;
+            let amount = if let Some(val) = validator {
+                u128::from(val.stake) + amount as u128
+            } else {
+                amount as u128
+            };
```
mina86, line 115:
```suggestion
            let amount = validator.map_or(0, |val| u128::from(val.stake)) + u128::from(amount);
```
should work, no?

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -161,6 +180,29 @@ pub mod restaking {
...
+        // Call Guest chain to update the stake
+        let validator_key = match service {
+            Service::GuestChain { validator } => validator,
+        };
+        let validator = chain
+            .candidate(*validator_key)
+            .map_err(|_| ErrorCodes::OperationNotAllowed)?
+            .ok_or(ErrorCodes::AccountValidationFailedForCPI)?;
+        let validator_stake = u128::from(validator.stake)
+            .checked_sub(amount as u128)
```
mina86, line 192:
```suggestion
            .checked_sub(u128::from(amount))
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -341,23 +381,42 @@ pub mod restaking {
...
+            solana_ibc::chain::ChainData::try_deserialize(&mut chain_data)
+                .unwrap();
+        let validator = chain
+            .validator(validator_key)
+            .map_err(|_| ErrorCodes::OperationNotAllowed)?;
+        let amount = if let Some(val) = validator {
+            u128::from(val.stake) + amount as u128
+        } else {
+            amount as u128
+        };
```
mina86, line 400: As above.

mina86 (approved):

solana/restaking/programs/restaking/src/lib.rs
dhruvja: yes borrow is enough.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: yes i guess it would.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: lazy evaluation sounds better to me.

dhruvja: clippy seems to say the lazy evaluation is unneccesary so we can keep it as it is.

PR #218, opened 2024-02-15
validator: sign pending blocks

dhruvja:
Checks and signs pending block if present before it listens to the events since until the pending block is signed, no new block event would be emitted.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -60,7 +60,7 @@ pub fn run_validator(config: Config) {
                 .get(&validator.pubkey().into())
                 .is_none()
             {
-                log::info!("it has a pending block");
+                log::info!("Found Pending block");
                 let fingerprint = &chain_account
                     .pending_block()
                     .unwrap()
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -78,7 +78,10 @@ pub fn run_validator(config: Config) {
                 );
                 match tx {
                     Ok(tx) => {
-                        log::info!("Pending Block signed -> Transaction: {}", tx);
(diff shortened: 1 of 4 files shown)


validator/src/validator.rs
```diff
@@ -50,6 +49,51 @@ pub fn run_validator(config: Config) {
...
+                            tx
+                        );
+                        break;
+                    }
+                    Err(err) => {
+                        log::error!("Failed to send the transaction {err}")
+                    }
+                }
+            } else {
+                log::info!("You have already signed the pending block");
```
blasrodri, line 92: Don't log saying "you.."
"There are ..."

dhruvja: oh will do

validator/src/validator.rs
```diff
@@ -50,6 +49,51 @@ pub fn run_validator(config: Config) {
         .expect("Invalid Genesis Hash");
 
     loop {
+        // Check if there is a pending block to sign
+        let chain_account: ChainData = program.account(chain).unwrap();
+        if chain_account.pending_block().unwrap().is_some() {
```
mina86, line 54: As it stands, we’ll keep trying to sign this block over and over until it gets finalised. We should at least remember height of the last signed block and compare it with pending.

dhruvja: it checks for pending block and the signature. If it contains the signature it doesnt sign the block again and then moves on to listen to `new_block` events.

dhruvja: but yes, we need to keep record of last signed block.

mina86 (approved):

PR #221, opened 2024-02-16
write-account: add Free operation

mina86:
Introduce Free operation to the write-account program which allows the
Write account to be freed and all lamports stored within it
transferred back to the Payer.

### DIFF
--- a/solana/write-account/src/instruction.rs
+++ b/solana/write-account/src/instruction.rs
@@ -1,37 +1,8 @@
-use solana_program::instruction::AccountMeta;
+use solana_program::instruction::{AccountMeta, Instruction};
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::Pubkey;
 
 type Result<T = (), E = ProgramError> = core::result::Result<T, E>;
-
-/// Generates instruction data for a Write operation.
-///
-/// `seed` and `bump` specifies seed and bump of the Write PDA.  Note that the
-/// actual seed used to create the PDA is `[payer.key, seed]` rather than just
-/// `seed`.  Use [`write_find_account`] to figure out the correct Write account
-/// with bump and generate the instruction data.
-///
-/// `offset` specifies offset in the write account to write data at and `data`
-/// is the data to write.
-pub fn write(
-    seed: &[u8],
-    bump: u8,
-    offset: usize,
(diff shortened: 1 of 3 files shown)


solana/write-account/src/program.rs
```diff
@@ -136,63 +110,139 @@ fn handle_write(
...
+        // Write account.  Must be writable and PDA.
+        let write = next_account_info(accounts)?;
+        if !write.is_writable {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        let seed_len = read(instruction, u8::from_le_bytes)?;
+        let seed_and_bump = read_slice(instruction, seed_len as usize + 1)?;
+        let this = Self { payer, write, seed_and_bump };
+
+        match Pubkey::create_program_address(&this.write_seeds(), program_id) {
```
dhruvja, line 233: Since we are verifying the write program seeds here, we should skip this check again in `setup_write_program` method right? I feel we should also verify bumps here.
https://github.com/ComposableFi/emulated-light-client/blob/f7f64888b41b53f70ed068be96a944577e50ad11/solana/write-account/src/program.rs#L149-L152

solana/write-account/src/instruction.rs
```diff
@@ -147,6 +123,45 @@ impl core::iter::Iterator for WriteIter<'_> {
...
+/// If `write_account` is not given, it’s going to be generated from provided
+/// Write program id, Payer account, seed and bump.
+pub fn free(
+    write_program_id: Pubkey,
+    payer: Pubkey,
+    write_account: Option<Pubkey>,
+    seed: &[u8],
+    bump: u8,
+) -> Result<Instruction> {
+    let mut buf = [0; { solana_program::pubkey::MAX_SEED_LEN + 3 }];
```
blasrodri, line 141: what's the three?

blasrodri: oh, I get it

solana/write-account/src/program.rs
mina86: The check in setup_write_account has indeed been removed.

dhruvja: oh sorry, my bad.

dhruvja (approved):

PR #227, opened 2024-02-19
solana-ibc: display pubkey using base58 instead of base64

dhruvja:
Since solana uses base58 to display public keys, its better to use the same.

### DIFF
--- a/solana/ed25519/src/lib.rs
+++ b/solana/ed25519/src/lib.rs
@@ -239,20 +239,20 @@ impl SignatureOffsets {
 }
 
 macro_rules! fmt_impl {
-    (impl $trait:ident for $ty:ident) => {
+    (impl $trait:ident for $ty:ident, $func_name:ident) => {
         impl fmt::$trait for $ty {
             #[inline]
             fn fmt(&self, fmtr: &mut fmt::Formatter) -> fmt::Result {
-                base64_display(&self.0, fmtr)
+                $func_name(&self.0, fmtr)
             }
         }
     };
 }
 
-fmt_impl!(impl Display for PubKey);
-fmt_impl!(impl Debug for PubKey);
-fmt_impl!(impl Display for Signature);
-fmt_impl!(impl Debug for Signature);
+fmt_impl!(impl Display for PubKey, base58_display);
+fmt_impl!(impl Debug for PubKey, base58_display);
+fmt_impl!(impl Display for Signature, base64_display);
+fmt_impl!(impl Debug for Signature, base64_display);
 
 /// Displays slice using base64 encoding.  Slice must be at most 64 bytes
 /// (i.e. length of a signature).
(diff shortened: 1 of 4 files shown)


solana/ed25519/src/lib.rs
```diff
@@ -266,6 +266,13 @@ fn base64_display(bytes: &[u8], fmtr: &mut fmt::Formatter) -> fmt::Result {
     fmtr.write_str(unsafe { core::str::from_utf8_unchecked(&buf[..len]) })
 }
 
+/// Displays slice using base58 encoding.
+fn base58_display(bytes: &[u8], fmtr: &mut fmt::Formatter) -> fmt::Result {
+    let data = bs58::encode(bytes).into_string();
+    // SAFETY: base58 fills the buffer with ASCII characters only.
```
mina86, line 272: There’s no `unsafe` code so there’s no need for `SAFETY` comment.
Alternative is to actually use some unsafe to optimise things a bit:

```rust
/// Displays slice using base58 encoding.
// TODO(mina86): Get rid of this once bs58 has this feature.  There’s currently
// PR for that: https://github.com/Nullus157/bs58-rs/pull/97
fn base58_display(bytes: &[u8], fmtr: &mut fmt::Formatter) -> fmt::Result {
    debug_assert!(self.0.len() <= 65);
    // The largest buffer we’re ever encoding is 65-byte long.  Base58
    // increases size of the value by less than 40%.  96-byte buffer is
    // therefore enough to fit the largest value we’re ever encoding.
    let mut buf = [0u8; 96];
    let len = bs58::encode(self.0).into(&mut buf[..]).unwrap();
    let output = &buf[..len];
    // SAFETY: We know that alphabet can only include ASCII characters
    // thus our result is an ASCII string.
    fmtr.write_str(unsafe { std::str::from_utf8_unchecked(output) })
}
```

solana/ed25519/src/lib.rs
```diff
@@ -266,6 +266,13 @@ fn base64_display(bytes: &[u8], fmtr: &mut fmt::Formatter) -> fmt::Result {
     fmtr.write_str(unsafe { core::str::from_utf8_unchecked(&buf[..len]) })
 }
 
+/// Displays slice using base58 encoding.
+fn base58_display(bytes: &[u8], fmtr: &mut fmt::Formatter) -> fmt::Result {
```
mina86, line 270: Perhaps `bytes: &[u8; 32]` and similarly in base64_display change to `bytes: &[u8; 64]`?

mina86 (approved):

solana/ed25519/src/lib.rs
dhruvja: oh i see that makes sense. I was wondering why it was the same in `base64_display` method. Got it now.

solana/ed25519/src/lib.rs
dhruvja: sounds good.

solana/ed25519/src/lib.rs
mina86: In `base64_display` there was a SAFETY comment because of the `unsafe` block wrapping `core::str::from_utf8_unchecked` call.

dhruvja: gotcha

PR #228, opened 2024-02-23
solana-ibc: make the inner value of map as public

dhruvja:
Making inner value of map as public so that we can use the associative function of the `LinearMap` which is needed on the relayer for fetching the value from the map for a given key and so on.


mina86:
Why is derive_more::Deref and derive_more::DerefMut not sufficient?

dhruvja:
Sorry i feel i missed that. I guess `deref` and `deref_mut` are sufficient.

PR #229, opened 2024-02-25
feat: implementation of cross chain calls specification README.md

dzmitry-lahoda:
@mina86 @dhruvja @blasrodri  please take a look. alignment on this is critical to start development of solana cross chain calls (including cvm calls).

### DIFF
--- a/solana/ics27/README.md
+++ b/solana/ics27/README.md
@@ -1,0 +1,19 @@
+
+Program implementating cross chain Programs calls ICS27 specification.
+
+Program works that way:
+
+1. Relayer calls `deliver` on Program with packet data.
+2. Program checks that packet was proved by IBC core Program account.
+3. Program parses packet to standard and verifies that accounts provided by relayer are accounts in packets
+4. Program executes cross chain transaction adhering all spec requirements
+5. Program ICS version will be `svm/borsh-1`.
+6. Program calls ibc core program `confirm` with ACK or FAIL status
+7. IBC core program moves packet to relevant status. It is up to ibc core program to verify TIMEOUT, port-caller Program mapping.
+
+Simple programs will be executed. 
+
+Not part of first implementation are next feature (these for later development):
+
+- ability of ICS-27 Program to request state of other packets, specifically ICS20 and ICS721 and ICS27 packets success status, by sequence id
+- for Program calls with many accounts, use account lookup tables to compress accounts  
--- a/solana/ics27/README.md
+++ b/solana/ics27/README.md
@@ -1,19 +1,37 @@
+# Program Implementing Cross-Chain Calls Based on ICS27 Specification
 
-Program implementating cross chain Programs calls ICS27 specification.
+This document outlines the operation of a program designed to facilitate cross-chain calls in adherence to the ICS27 specification. The program enables the execution of cross-chain transactions, leveraging the Inter-Blockchain Communication (IBC) protocol for secure and verified interactions between different blockchain networks.
 
-Program works that way:
+## Workflow:
 
-1. Relayer calls `deliver` on Program with packet data.
-2. Program checks that packet was proved by IBC core Program account.
-3. Program parses packet to standard and verifies that accounts provided by relayer are accounts in packets
-4. Program executes cross chain transaction adhering all spec requirements
-5. Program ICS version will be `svm/borsh-1`.
-6. Program calls ibc core program `confirm` with ACK or FAIL status
-7. IBC core program moves packet to relevant status. It is up to ibc core program to verify TIMEOUT, port-caller Program mapping.
+### 1. Relayer Interaction:
+- A relayer, an off-chain entity, invokes the `deliver` function on the program, passing in packet data for processing.
 
-Simple programs will be executed. 
+### 2. Packet Verification:
+- The program verifies that the packet has been authenticated by the IBC core program's account, ensuring its legitimacy.
 
-Not part of first implementation are next feature (these for later development):
+### 3. Packet Parsing and Account Verification:
+- It parses the packet according to the `svm/borsh-1` standard, ensuring the packet data is correctly serialized/deserialized.
+- Verifies that the accounts specified by the relayer match those included in the packet, ensuring consistency and authenticity.
 
-- ability of ICS-27 Program to request state of other packets, specifically ICS20 and ICS721 and ICS27 packets success status, by sequence id
-- for Program calls with many accounts, use account lookup tables to compress accounts  
+### 4. Execution of Cross-Chain Transaction:
+- Executes the transaction detailed in the packet, adhering to all specifications and requirements mandated by the ICS27 protocol.
+
+### 5. ICS Version Compatibility:
+- The program operates under the `svm/borsh-1` ICS version, ensuring compatibility with the serialization and deserialization standards.
+
+### 6. Acknowledgment Handling:
+- Post-execution, the program communicates with the IBC core program, conveying an ACK (acknowledgment) or FAIL status to indicate the outcome of the transaction.
+
+### 7. IBC Core Program Coordination:
+- The IBC core program is responsible for updating the packet's status based on the received acknowledgment. It manages TIMEOUT scenarios and verifies the port-caller program mapping, ensuring proper transaction flow and integrity.
+
+## Exclusions from Initial Implementation:
+
+The first version of the program will not include:
+- The capability to query the state of other packets, such as ICS20, ICS721, and ICS27 packets, by sequence ID, a feature commonly utilized in cosmos-ibc and by contracts in Osmosis WASM hooks protocols and CVM on Cosmos-CW.
+- The use of Address Translation Tables (ALTs) for compressing account information in transactions involving multiple accounts. Future versions may explore packet instructions for creating/extending LUTs or creating ALTs through ICS27 owner/IBC core governance/CVM owner initiatives for reuse in any packets later.
+
+Said that CVM is no operation without implementing both, and heavy realie on IBC relayer and IBC core impementation to support above flows and enabling CVM.
+
+This specification delineates the functional blueprint for a program enabling cross-chain transactions via the ICS27 protocol, specifying the workflow, requirements, and current limitations while acknowledging potential areas for future development.
--- a/solana/ics27/README.md
+++ b/solana/ics27/README.md
@@ -4,7 +4,7 @@ This document outlines the operation of a program designed to facilitate cross-c
 
 ## Workflow:
 
-### 1. Relayer Interaction:
+### 1 Relayer Interaction
 
 
 A Relayer sends full IBC packet(header+data) into `simulate` instruction into contract.
@@ -22,6 +22,7 @@ Extend lookup table can be output several times during simulation, relayer shoul
 { 
 "ExtendLookupTable" : { 
   new_addresses: ["pubkey1", .., "pubkeyN"] 
+ }
 }
 ```
 
--- a/solana/ics27/README.md
+++ b/solana/ics27/README.md
@@ -5,51 +5,89 @@ This document outlines the operation of a program designed to facilitate cross-c
 ## Workflow:
 
 ### 1. Relayer Interaction:
-- A relayer, an off-chain entity, invokes the `deliver` function on the program, passing in packet data for processing.
+
+
+A Relayer sends full IBC packet(header+data) into `simulate` instruction into contract.
+
+Contract outputs all accounts it needs for execution as **Anchor compatible event**.
+
+Simulate method can output more arbitrary logs and evnets, relayer ignores them.
+
+Simulate method can fail TX, but as long as log was output relayer should try relay with these accounts.
+
+Extend lookup table can be output several times during simulation, relayer should use all of them.
+
+```json
+// naming used to adhere that event is command for next step in flow
+{ 
+"ExtendLookupTable" : { 
+  new_addresses: ["pubkey1", .., "pubkeyN"] 
+}
+```
+
+A Relayer invokes the `deliver` function on the program, passing in packet data and accounts for processing.
 
 ### 2. Packet Verification:
-- The program verifies that the packet has been authenticated by the IBC core program's account, ensuring its legitimacy.
+
+The program verifies that the packet has been authenticated by the IBC core program's account, ensuring its legitimacy.
 
 ### 3. Packet Parsing and Account Verification:
-- It parses the packet according to the `svm/borsh-1` standard, ensuring the packet data is correctly serialized/deserialized.
-- Verifies that the accounts specified by the relayer match those included in the packet, ensuring consistency and authenticity.
+
+It parses the packet according to the `svm/borsh-1` standard, ensuring the packet data is correctly serialized/deserialized.
+Verifies that the accounts specified by the relayer match those included in the packet, ensuring consistency and authenticity.
 
 ### 4. Execution of Cross-Chain Transaction:
-- Executes the transaction detailed in the packet, adhering to all specifications and requirements mandated by the ICS27 protocol.
+
+Executes the transaction detailed in the packet, adhering to all specifications and requirements mandated by the ICS27 protocol.
 
 ### 5. ICS Version Compatibility:
-- The program operates under the `svm/borsh-1` ICS version, ensuring compatibility with the serialization and deserialization standards.
+
+The program operates under the `svm/borsh-1` ICS version, ensuring compatibility with the serialization and deserialization standards.
+Where `borsch` is encoding, `svm` is transaction type and `1` is version of implementation. 
 
 ### 6. Acknowledgment Handling:
-- Post-execution, the program communicates with the IBC core program, conveying an ACK (acknowledgment) or FAIL status to indicate the outcome of the transaction.
+
+Post-execution, the program communicates with the IBC core program, conveying an ACK (acknowledgment) or FAIL status to indicate the outcome of the transaction.
 
 ### 7. IBC Core Program Coordination:
-- The IBC core program is responsible for updating the packet's status based on the received acknowledgment. It manages TIMEOUT scenarios and verifies the port-caller program mapping, ensuring proper transaction flow and integrity.
 
-## Exclusions from Initial Implementation:
+The IBC core program is responsible for updating the packet's status based on the received acknowledgment. It manages TIMEOUT scenarios and verifies the port-caller program mapping, ensuring proper transaction flow and integrity.
+
+## Extensions of initial implementation:
 
 The first version of the program will not include:
 - The capability to query the state of other packets, such as ICS20, ICS721, and ICS27 packets, by sequence ID, a feature commonly utilized in cosmos-ibc and by contracts in Osmosis WASM hooks protocols and CVM on Cosmos-CW.
-- The use of Address Translation Tables (ALTs) for compressing account information in transactions involving multiple accounts. Future versions may explore packet instructions for creating/extending LUTs or creating ALTs through ICS27 owner/IBC core governance/CVM owner initiatives for reuse in any packets later.
+- ALTs for compressing account information in transactions involving multiple accounts. Future versions may explore packet instructions for creating/extending LUTs or creating ALTs through ICS27 owner/IBC core governance/CVM owner initiatives for reuse in any packets later.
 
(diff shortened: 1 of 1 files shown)


solana/ics27/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Workflow:
+
+### 1. Relayer Interaction:
+- A relayer, an off-chain entity, invokes the `deliver` function on the program, passing in packet data for processing.
+
+### 2. Packet Verification:
+- The program verifies that the packet has been authenticated by the IBC core program's account, ensuring its legitimacy.
+
+### 3. Packet Parsing and Account Verification:
+- It parses the packet according to the `svm/borsh-1` standard, ensuring the packet data is correctly serialized/deserialized.
```
mina86, line 14: What’s svm/borsh-1?

solana/ics27/README.md
```diff
@@ -0,0 +1,55 @@
...
+- Post-execution, the program communicates with the IBC core program, conveying an ACK (acknowledgment) or FAIL status to indicate the outcome of the transaction.
+
+### 7. IBC Core Program Coordination:
+- The IBC core program is responsible for updating the packet's status based on the received acknowledgment. It manages TIMEOUT scenarios and verifies the port-caller program mapping, ensuring proper transaction flow and integrity.
+
+## Exclusions from Initial Implementation:
+
+The first version of the program will not include:
+- The capability to query the state of other packets, such as ICS20, ICS721, and ICS27 packets, by sequence ID, a feature commonly utilized in cosmos-ibc and by contracts in Osmosis WASM hooks protocols and CVM on Cosmos-CW.
+- The use of Address Translation Tables (ALTs) for compressing account information in transactions involving multiple accounts. Future versions may explore packet instructions for creating/extending LUTs or creating ALTs through ICS27 owner/IBC core governance/CVM owner initiatives for reuse in any packets later.
```
mina86, line 33: If by ALT in this context you mean address lookup table from Solana’s versioned transactions that this isn’t exactly true. The off-chain component which triggers execution of the hook/CVM program will most likely need to use them.  As you’ve pointed out some operations may require over 30 accounts and an *absolute* limit of accounts in legacy transaction is 34 (and with that limit you have like 9 bytes to spare).

solana/ics27/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Said that CVM is no operation without implementing both, and heavy realie on IBC relayer and IBC core impementation to support above flows and enabling CVM.
+
+This specification delineates the functional blueprint for a program enabling cross-chain transactions via the ICS27 protocol, specifying the workflow, requirements, and current limitations while acknowledging potential areas for future development.
+
+## Anchor usage
+
+I am not sure of anchor usage. I actutually never coded Anchor. If Achor exprert will 
+As of now it has (after 3 years in dev) limitation - reason why Michal forked Anchor and not supporting some scenarios like we dicussing https://forum.solana.com/t/srfc-00002-off-chain-instruction-account-resolution/25/3
```
mina86, line 42: The reason for fork is really inconsequential. We carry only a single-line diff and it has to do with how Anchor smart contract defines entrypoint. We wanted to be able to hook in before Anchor’s entry function is called.

As far as I understand the main problem with Anchor would be calling the contract though I admit that I’m not impressed by Anchor and wouldn’t mind if ICS20/ICS27/CVM-related code would live in its own contract which wouldn’t use Anchor.

blasrodri:
My understanding is that we choose ICS27 in favor of ICS20+memo. The question is, how do we operate with asset transfers? Because operations might need assets to be moved from one domain to another. I understand that this involves more than one transaction, and that's ok. 
However, we need to specify on the protocol how we operate with regards to these asset transfers, upon failure and success.

solana/ics27/README.md
dzmitry-lahoda: will update do. but this is 
```
ICS-27 takes advantage of [ICS-04 channel version negotiation](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md#versioning) to negotiate metadata and channel parameters during the channel handshake. The metadata will contain the encoding format along with the transaction type so that the counterparties can agree on the structure and encoding of the interchain transactions.
```
So encoding is borsh, transactions type is `svm` (it can be neon-evm on solana for example) and `1` is version in case we need to bump it.

See ics20 version example.

solana/ics27/README.md
dzmitry-lahoda: ALT is official short cut of Solana for Address Lookup Tables

dzmitry-lahoda: As I read the code, program is not aware if it was called with ALT or not. This is sender only knowledge.

dzmitry-lahoda: By future I mean after simple no ALT transactions tested e2e. So some bugs in relayer creating ALTs will not hinder progress.

solana/ics27/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+So I do not plan to use Anchor to avoid issues.
```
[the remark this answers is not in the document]
dzmitry-lahoda: we have fork of anchor, means anchor does not supports whole solana. also see please link - so anchor does not supports even more of solana. also see github issues on anchor - for last years anchor either was lagging behind solana in features or never implemented. 

also i doubt anchor supports non borsh encoding if ever it will be good (protobuf for example may appear be better for CVM native transactions).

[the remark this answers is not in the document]
dzmitry-lahoda: @blasrodri please look at whole text. link one of example https://forum.solana.com/t/srfc-00002-off-chain-instruction-account-resolution/25/3 which exactly discussed simulation/accounts abstractions vs anchor.

dzmitry-lahoda:
Will update docorz with this. But I will not implement first version with this. Need to do e2e tests of just calls, no ALTs no ICS20 checks.

dzmitry-lahoda:
for interface amid relayer and solana-ics27 contract.
- relayer sends full IBC packet(header+data) into `simulate instruction` into contract.
- contract outputs all accounts it needs for execution as JSON (JSON shape defined as part of solana crates shared amid relayer and solana contract).
- simulate method can output more arbitrary logs other than that JSON, it is up to relayer to find relevant log.
- simulate method can fails TX, so as long as log was output relayer should try relay with these accounts.
like this may be
- extend lookup table can be output several times during simulation, relayer should use all of them.  because CVM program tranfer is recursive, will output for each instuction in tree.

```json
{ 
"ExtendLookupTable" : { 
  new_addresses: ["pubkey1", .., "pubkeyN"] 
}
```

mina86:
> contract outputs all accounts it needs for execution as JSON (JSON shape defined as part of solana crates shared amid relayer and solana contract).

This probably should be Borsh.  Solana is all about Borsh.  The contract would define an Event type (probably an enum) which would be used to output information including the accounts.

> simulate method can output more arbitrary logs other than that JSON, it is up to relayer to find relevant log.

Right, that’s typical of Solana as far as I understand.

dzmitry-lahoda:
> This probably should be Borsh. Solana is all about Borsh. The contract would define an Event type (probably an enum) which would be used to output information including the accounts.

I want easy way to see logs visually for debugging. I do not plan develop relayer debugging tools or run my relayer. Binary implies people doing relayer care logging, but that was never the case with hyperspace. So I do what will work for sure.

mina86:
That’s more to do with how things on Solana are done. It also makes it easier to separate debug logs coming from the contract and events emitted by the contract.

dzmitry-lahoda:
in solana all events are logs 

https://solana.stackexchange.com/questions/2064/anchor-events-what-why-how

i want accounts to be tagged JSON arrays for  ease of debugging. other easy simple and cheap options to debug and at same time provide structure to relayr? 

May be instead of `simulate`, there should be `create_accounts_table/extend_account_table` in program?

dzmitry-lahoda:
can in theory reuse anchor events declaration/format  logging, if it can be easy reused without running their full infra/cli and using 100% of it.

mina86:
> in solana all events are logs

Kind of.  There’s a [`sol_log`](https://docs.rs/solana-program/latest/solana_program/log/fn.sol_log.html) host function which takes a string argument and [`sol_log_data`](https://docs.rs/solana-program/latest/solana_program/log/fn.sol_log_data.html) which takes bytes slices.  While they are both logs, the are logged differently.  For example, here’s partial output from our anchor tests:

    Transaction executed in slot 32:
      Signature: 28Le6qesM7znSsAKnpQ2dVEL6deWiVkjmArxhF7QmZiscxvmJaRRwXxch9MKvMuwQG7xGE2DKyjuS8CS35RRKJQP
      Status: Ok
      Log Messages:
        Program 8J9Tc5fdkt1aqGuFwpR92MCPx8Gr5KP7WEm5rvYbDqis invoke [1]
        Program log: Instruction: Initialise
        Program 11111111111111111111111111111111 invoke [2]
        Program 11111111111111111111111111111111 success
        Program 11111111111111111111111111111111 invoke [2]
        Program 11111111111111111111111111111111 success
        Program 11111111111111111111111111111111 invoke [2]
        Program 11111111111111111111111111111111 success
        Program data: AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAIAAAAAAAAAAAVhUFe1K0FwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA>
        Program 8J9Tc5fdkt1aqGuFwpR92MCPx8Gr5KP7WEm5rvYbDqis consumed 30414 of 200000 compute units
        Program 8J9Tc5fdkt1aqGuFwpR92MCPx8Gr5KP7WEm5rvYbDqis success

Observe `Program log:` lines are human-readable logs while `Program data:` lines hold base64-encoded events.  This difference is what allows clients to distinguish between debug logs and events.  When contracts emit events they are emitted as data and later presented encoded with base64.

mina86:
> can in theory reuse anchor events declaration/format logging, if it can be easy reused without running their full infra/cli and using 100% of it.

Anchor doesn’t build much on top of plain Solana’s logging.  What anchor does is add eight-byte discriminant to every event and when events are emitted prefixes Borsh-serialised event with that discriminant.  Clients can then use the first eight bytes to figure out type of the event.

In solana-ibc we’re not actually using Anchor’s mechanism.  Instead we have an Event enum which is Borsh-serialisable and output that directly.  That’s the approach I’d recommend rather than trying to incorporate parts of Anchor.

dzmitry-lahoda:
thanks about explanation of data.

> In solana-ibc we’re not actually using Anchor’s mechanism. Instead we have an Event enum which is Borsh-serialisable and output that directly. That’s the approach I’d recommend rather than trying to incorporate parts of Anchor.

No way. There are 2 wasy, JSON and Anchor. You invented 3rd.... Why ? 

As of anchor, chances indexers will get it same for JSON. For custom events I doubt.

mina86:
It’s not a third way.  It’s just Borsh which is *very* commonly used on Solana, there’s nothing special about it.  With Borsh all you need to use the events is the file with the event type definition.  With Anchor you now need to pull in Anchor dependency.  

Not everything in Solana uses Anchor so I don’t share your pessimism that indexers wouldn’t support Borsh.  Nonetheless, I don’t know what indexer we’re talking about so cannot help there.

dzmitry-lahoda:
https://github.com/coral-xyz/anchor/blob/master/tests/events/programs/events/src/lib.rs

    anchor_lang::solana_program::log::sol_log_data(&[&anchor_lang::Event::data(&#data)]);

https://github.com/coral-xyz/anchor/blob/master/lang/src/lib.rs#L292


https://github.com/coral-xyz/anchor/blob/23028334f3ef1a9e309a7d9ab552fd218ca2a00d/lang/attribute/event/src/lib.rs#L43C45-L43C58

dzmitry-lahoda:
> Not everything in Solana uses Anchor so I don’t share your pessimism that indexers wouldn’t support Borsh. Nonetheless, I don’t know what indexer we’re talking about so cannot help there.

I see anchor does prefix to data, and I can wrap parts of the code for anchor to handle indexing.

@mina86 I you will consider handling part of responsibility for doing production debugging in area of cross chain calls and relaying, I would happy to listen to you. but from my perspective only JSON or Anchro.

here is Anchor

https://github.com/aleph-im/pyaleph

dzmitry-lahoda:
@dhruvja will use just list of raw IX in packet data data.

dhruvja:
> > Not everything in Solana uses Anchor so I don’t share your pessimism that indexers wouldn’t support Borsh. Nonetheless, I don’t know what indexer we’re talking about so cannot help there.
> 
> I see anchor does prefix to data, and I can wrap parts of the code for anchor to handle indexing.
> 
> @mina86 I you will consider handling part of responsibility for doing production debugging in area of cross chain calls and relaying, I would happy to listen to you. but from my perspective only JSON or Anchro.
> 
> here is Anchor
> 
> https://github.com/aleph-im/pyaleph
> 
> https://github.com/aleph-im/aleph-indexer-library
> 
> https://github.com/aleph-im/aleph-indexer-library/blob/main/packages/indexer-generator/README.md
> 
> I am not fun of anchor for doing everything, but data integration is good for me.

Well the indexer library only uses anchor since it provides IDL. But if you write the indexer in rust, you can just use the interfaces from the contract and serialize and deserialize it. I dont see why else we would need anchor?

dzmitry-lahoda:
I will split out new spec. here will retain only relevant API with relayer and ibc core.

dzmitry-lahoda:
@dhruvja did you looked at links? it has indexer.

dzmitry-lahoda:
@dhruvja please read other comments? events to be in JSON(indexer exists defacto in any database base decode and pare here is) or in something with exisitng indexer.

i am ok with JSON.

dzmitry-lahoda:
@blasrodri I will need observe how solana is going on. I am going to use JSON or Anchro format for events as these already have good indexers. I will be blocked without. Do you wish me to waste time writing custom indexer?

dzmitry-lahoda:
@dhruvja I want to be able to get indexer for any events from cross chain calls. anchor_lang gives that ability because it is integrated with generator for events and state. anchor_lang already used by you. I will use it just for doing data part. 

Anyway. I will not code ics27/cvm in this repo. will so just sent new spec for ibc modules so that can attach external contract to handle port/channels so we do not argue on some non.

dzmitry-lahoda:
> yes i looked at those. Well seems like i am kinda lost now, what events do you want to index. Ibc events?

all events and data stoge in my contract will be in acnhor format. i will use it.

dzmitry-lahoda:
@dhruvja do you have any objection using Anchor events format which will be outputed for all events  from what will be cross chain call contract? What are these?  From michal  I have heard size. But as I see he is using anchor crate in ibc core. So do not see what problem.

What else?

Using anchor for declaritng state and events is of huge utilitity - it gives indexer for free.  Not sure if few kilobytes of contract size worth that suffer.

dzmitry-lahoda:
@mina86 what are objections using anchor events format?

dhruvja:
> @dhruvja do you have any objection using Anchor events format which will be outputed for all events from what will be cross chain call contract? What are these? From michal I have heard size. But as I see he is using anchor crate in ibc core. So do not see what problem.
> 
> What else?
> 
> Using anchor for declaritng state and events is of huge utilitity - it gives indexer for free. Not sure if few kilobytes of contract size worth that suffer.

we skipped using anchor events because we were on the verge of overflowing heap. But then we found that we could increase heap size to 256kb ( previously it was 32kb ) though it needed custom allocator for it. I dont have any objection tbh. The size of the contract is not the issue, the issue is running out of stack or heap which is pretty limited in the runtime.
 .

dzmitry-lahoda:
> we skipped using anchor events because we were on the verge of overflowing heap

sorry, how it could be? Anchor events are just Borsch + some prefix name. See links. + build.rs time generator for interface. I am sure Anchor was no a problem.

dzmitry-lahoda:
>  The size of the contract is not the issue, the issue is running out of stack or heap which is pretty limited in the runtime.

Contract heap/stack will be small. If that is risk, than must not use borsh either (but use direct stream write). But afaik Michal did not proposed not using borsh.

dhruvja:
> > we skipped using anchor events because we were on the verge of overflowing heap
> 
> sorry, how it could be? Anchor events are just Borsch + some prefix name. See links. + build.rs time generator for interface. I am sure Anchor was no a problem.

well i think we used solana logging for events because for emitting events on anchor, we need to borsh serialize the event and then emit it which resulted in serializing it twice.

dhruvja:
> > The size of the contract is not the issue, the issue is running out of stack or heap which is pretty limited in the runtime.
> 
> Contract heap/stack will be small. If that is risk, than must not use borsh either (but use direct stream write). But afaik Michal did not proposed not using borsh.

stack is limited, so if we run out we need to move them to heap. But heap has can be increased till 256kb so i dont think that poses a risk

dzmitry-lahoda:
> well i think we used solana logging for events because for emitting events on anchor, we need to borsh serialize the event and then emit it which resulted in serializing it twice.

one can emit Anchor events without heap and without anchor if one wishes, and yet get benefit from IDL (like polkadot-sdk does).

I problem will araise, can easily fix it.

@dhruvja cross chain calls will be separate contract not part of this repo nor part of core ibc prove system. so we can avoid conustion and opinions on details out of this repo. currently I am writing new doc which just outlines interface of ibc-core for  ibc-apps

mina86:
Potential downside is that you do need some Anchor dependency or hard coding discriminants in addition to using Borsh if events are using Anchor format (IDL isn’t self-describing as far as I understand an in fact solana-ibc generates IDC for our events).  Also note that events which output the accounts that are needed are generated from simulation so those won’t exist on chain.  

> sorry, how it could be? Anchor events are just Borsch + some prefix name. See links. + build.rs time generator for interface. I am sure Anchor was no a problem.

Anchor was in fact the problem because Anchor isn’t well optimised.

> Contract heap/stack will be small. If that is risk, than must not use borsh either (but use direct stream write). But afaik Michal did not proposed not using borsh.

That’s not exactly true simply because Solana doesn’t support it.  Borsh in fact uses stream write but because logs in Solana have to be done through a single call to sol_log_data serialised object has to be accumulated in memory.  Just using streaming write doesn’t solve the problem.

dzmitry-lahoda:
> (IDL isn’t self-describing as far as I understand an in fact solana-ibc generates IDC for our events)

IDL is self describing, as it is allows to generate indexer, API, parsers, JSON mappings

dzmitry-lahoda:
> Anchor was in fact the problem because Anchor isn’t well optimised.

for me anchor is not problem. i writing new spec to avoid to be dragged in impl details of apps https://github.com/ComposableFi/emulated-light-client/pull/244/files#diff-5d753a01bdfd18da8fb3c735527ef633fc443c62325c7ce0f2bf47ccf6814278 . So let retain discussion only related to ibc-core integrations of apps. app internal details should be out of dicussion now.

dzmitry-lahoda:
> That’s not exactly true simply because Solana doesn’t support it. Borsh in fact uses stream write but because logs in Solana have to be done through a single call to sol_log_data serialised object has to be accumulated in memory. Just using streaming write doesn’t solve the problem.

not sure what you want to say with this. i do not see problem nor solution.

mina86:
> IDL is self describing, as it is allows to generate indexer, API, parsers, JSON mappings

I’m not sure this is true since to my knowledge it doesn’t include Borsh schema of objects being serialised.  I’m happy to be shown otherwise.

> for me anchor is not problem.

You’ve questioned whether Anchor caused lack of memory issues we’ve experienced.  It has.

> not sure what you want to say with this. i do not see problem nor solution.

You’ve made a statement about using streaming writes to reduce memory usage in a smart contract.  I’ve pointed out your suggestion isn’t as straightforward.

dzmitry-lahoda:
> I’m not sure this is true since to my knowledge it doesn’t include Borsh schema of objects being serialised. I’m happy to be shown otherwise.

I shown already. I would be happy to see opposite proposal showing how other schemas work better.

> You’ve made a statement about using streaming writes to reduce memory usage in a smart contract. I’ve pointed out your suggestion isn’t as straightforward.

it is easy. just preallocate slice for all events and stream into it. up to maximal limit. can split accounts by limit. do not see any problem.

dzmitry-lahoda:
here is requirements for ibc-core to support CVM https://github.com/ComposableFi/emulated-light-client/pull/244

PR #230, opened 2024-02-26
cf-guest: implement light client for the guest blockchain

mina86:
Add partial IBC light client implementation for the guest blockchain.
Misbehaviour handling isn’t yet done.


common/ibc-guest/proto/guest.proto
```diff
@@ -0,0 +1,56 @@
...
+
+	// Borsh-serialised epoch the block belongs to.  The epoch specifies
+	// validators which sign the block.
+	bytes epoch = 3;
+
+	repeated Signature signatures = 4;
+}
+
+message Signature {
+	uint64 index = 1;
```
blasrodri, line 49: what's this index?

common/ibc-guest/src/client.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn with_header(&self, header: &super::Header<PK>) -> Self {
+        let latest_height = core::cmp::max(
+            header.block_header.block_height,
+            self.latest_height,
+        );
+        let epoch_commitment = header
+            .block_header
+            .next_epoch_commitment
+            .as_ref()
+            .unwrap_or(&self.epoch_commitment)
```
blasrodri, line 42: can you comment when it's ok not to use next_epoch_commitment?

common/ibc-guest/proto/guest.proto
mina86: That’s the index in validator set defined in Epoch.

common/cf-guest/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() -> std::io::Result<()> {
+    prost_build::Config::new()
+        .enable_type_names()
+        .type_name_domain(["."], "composable.finance")
```
blasrodri, line 4: can we change to Picasso?

mina86: The idea is that the domain corresponds to an actual web domain. Though I guess sit wouldn’t be end of the world if we changed it to picasso.

PR #231, opened 2024-02-27
solana-ibc: Remove tendermint patch

dhruvja:
Tendermint patch was added to disable `flex-error/eyre_tracer` feature. But this change is merged (Ref: https://github.com/informalsystems/tendermint-rs/pull/1371 ). So we can safely remove the patch and use the latest version of tendermint.


dhruvja:
Seems like `ibc-rs` still uses version `0.34.0` which has `flex-error/eyre_tracer` feature enabled. So closing this until `ibc` uses `0.34.1`.

PR #240, opened 2024-03-06
solana-ibc: use sigverify program for checking Ed25519 signatures

mina86:
Introduce Verifier to solana-signature-verifier crate which provides
interface for checking Ed25519 signatures.  This replaces verifier
defined in solana-ed25519 crate with addition of supporting the
sigverify program.

This way, solana-ed25519 ends up only defining key and signature
types.  Rather than having those in a small crate, move it to
solana-signature-verifier as well.

With all that, update solana-ibc to use the new verifier.


dhruvja (approved):
LGTM

PR #246, opened 2024-03-12
guestchain: add option for generating blocks if head is too old

mina86:
Block generation is driven by changes to the state and to the set of validators.  As a consequence, it is theoretically possible for a new block to not be produced for a very long time.  Since blocks carry time information this means that clients may lack time information necessary to handle time-based events such as timeouts.

Add option for setting maximum block age after which a new block can be generated even if old block doesn’t change state nor finishes epoch.


dhruvja (approved):
LGTM

PR #251, opened 2024-03-20
cf-guest: provide explicit Any conversion methods and update type URLs

mina86:
To make it easier to convert to and from Any types which come from other crates than cf-guest depends on, provide AnyConvert trait with to_any and try_from_any methods which perform conversion between given type and Any message without relying on specific Any type.

Secondly, since IBC uses type URLs without domain in Any messages, do the same in our code base.  Firstly, change code converting from Any such that it ignores the domain in the type_url¹.  Secondly, replace TYPE_URL constant with IBC_TYPE_URL constant which excludes the domain.  And lastly, use that constant for GUEST_TYPE in AnyClientState and AnyConsensusState.

This new code will produce different Any representation for types. Though it will be capable of decoding old Any representations.

¹ This is in fact the correct behaviour.  The previous code which
  compared the domain was invalid according to Protocol Buffer
  specification.  So in addition to working better with IBC, this
  change also improves compatibility with Protobuf.

### DIFF
--- a/common/cf-guest/src/client_impls.rs
+++ b/common/cf-guest/src/client_impls.rs
@@ -3,9 +3,9 @@ use alloc::vec::Vec;
 
 use guestchain::PubKey;
 
+use super::proto::Any;
 use super::{
-    proof, Any, ClientMessage, ClientState, ConsensusState, Header,
-    Misbehaviour,
+    proof, ClientMessage, ClientState, ConsensusState, Header, Misbehaviour,
 };
 
 mod ibc {
--- a/common/cf-guest/src/lib.rs
+++ b/common/cf-guest/src/lib.rs
@@ -6,8 +6,6 @@ extern crate std;
 
 use alloc::string::ToString;
 
-use ibc_proto::google::protobuf::Any;
-
 mod client;
 mod client_impls;
 mod consensus;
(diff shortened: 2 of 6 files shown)


dhruvja:
if IBC doesnt use a type name, should we remove it as well? Then there wont be any confusion as such. What do you think?

mina86:
I’m not sure what you mean. IBC uses `/foo.bar.baz.Msg` as type URL. This PR changes URLs we’re using to match this format as well.

dhruvja:
we use a type domain giving us `type/foo.bar.baz.Msg` right? Instead we can just have `/type.foo.bar.baz.Msg`

mina86:
This PR changes that so that we no longer include the domain:
```rust
            pub const IBC_TYPE_URL: &'static str =
                concat!("/lightclients.guest.v1.", stringify!($Msg));
```

blasrodri (approved):

dhruvja:
ya but we could remove the domain from `build.rs` file as well right?

dhruvja (approved):

mina86:
> ya but we could remove the domain from `build.rs` file as well right?

Oh, I see what you mean.  Yeah, that might be easier.  I noticed your comment only after merging though.  I’ll take a look and potentially send a followup PR.

dhruvja:
sounds good no worries

PR #252, opened 2024-03-20
cf-guest: implement misbehaviour handling

mina86:
Add handling of misbehaviours.  Both detection of headers which imply a misbehaviour as well as handling of explicit Misbehaviour message.

As part of this, add support for accepting blocks from previous epoch. Previously, once we received a block which is the last block of an epoch we would only accept headers from the new epoch.  Now we continue to accept headers from current epoch.


common/cf-guest/src/client_impls.rs
```diff
@@ -369,26 +394,95 @@ impl<PK: PubKey> ClientState<PK> {
...
+                    if header_time_ns >= next_time_ns {
+                        return Ok(true);
+                    }
+                }
+                false
+            }
+        })
     }
 
     fn check_for_misbehaviour_misbehavior(
```
blasrodri, line 465: this name is a bit odd?

mina86: `check_for_misbehaviour` is the action and `misbehaviour` is type of client message.  I’ve changed this to include `in` as in `check_for_misbehaviour_in_misbehaviour` to maybe make it clearer?  It’s not a great name but it matches `check_for_misbehaviour_in_header` and I’m not inventive enough to think of something better. ;)

common/cf-guest/src/client_impls.rs
```diff
@@ -369,26 +394,95 @@ impl<PK: PubKey> ClientState<PK> {
...
+                existing_state != header_state
+            }
+
+            Neighbourhood::Neighbours(prev, next) => {
+                // Otherwise, make sure that timestamp of each consensus is
+                // strictly increasing.  If it isn’t, that’s evidence of
+                // misbehaviour.
+                let header_time_ns = header.block_header.timestamp_ns.get();
+                if let Some(prev_time_ns) = get_timestamp_ns(prev)? {
+                    if header_time_ns <= prev_time_ns {
```
blasrodri, line 451: can it be equal if it's the previous?

common/cf-guest/src/client_impls.rs
```diff
@@ -369,26 +394,95 @@ impl<PK: PubKey> ClientState<PK> {
...
+
+        // If blocks have the same height they must be the same (i.e. have the
+        // same hash).  Otherwise, check if their timestamp is strictly
+        // increasing.  If those conditions are not met we have a proof of
+        // misbehaviour.
+        let block1 = header1.block_header;
+        let block2 = header2.block_header;
+        Ok(match block1.block_height.cmp(&block2.block_height) {
+            Ordering::Less => block1.timestamp_ns >= block2.timestamp_ns,
+            Ordering::Equal => header1.block_hash != header2.block_hash,
```
blasrodri, line 483: ==?

common/cf-guest/src/client_impls.rs
mina86: No.  Timestamps are strictly increasing.  This is checked in Block::generate_next:

```rust
        } else if timestamp_ns <= self.timestamp_ns {
            return Err(GenerateError::BadHostTimestamp);
        }
```

(here `self` is the current block and `timestamp_ns` is timestamp of
the next block to be generated).

common/cf-guest/src/client_impls.rs
mina86: The condition here checks if the header is invalid.  That is, if the two block hashes aren’t equal, than there is proof of misbehaviour.

blasrodri (approved):
can we add some tests?

PR #254, opened 2024-03-20
solana-ibc: store protobuf encoded value in trie

dhruvja:
We used to store hash of the borsh encoded value in the trie but IBC encodes the value in protobuf and checks the proof against it. So encoding the value as protobuf so that we match the value which IBC expects.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
@@ -6,6 +6,8 @@ use crate::consensus_state::AnyConsensusState;
 use crate::ibc;
 use crate::storage::{self, IbcStorage};
 
+use ibc_proto::Protobuf;
+
 type Result<T = (), E = ibc::ContextError> = core::result::Result<T, E>;
 
 impl ibc::ClientExecutionContext for IbcStorage<'_, '_> {
@@ -21,7 +23,9 @@ impl ibc::ClientExecutionContext for IbcStorage<'_, '_> {
         msg!("store_client_state({}, {:?})", path, state);
         let mut store = self.borrow_mut();
         let mut client = store.private.client_mut(&path.0, true)?;
-        let hash = client.client_state.set(&state)?.digest_with_client(&path.0);
+        client.client_state.set(&state)?;
+        let state_any = state.encode_vec();
+        let hash = cf_guest::digest_with_client_id(&path.0, state_any.as_slice());
         let key = trie_ids::TrieKey::for_client_state(client.index);
         store.provable.set(&key, &hash).map_err(error)
     }
@@ -89,14 +93,16 @@ impl IbcStorage<'_, '_> {
             (head.timestamp_ns, head.block_height)
         };
 
+        let encoded_state = state.clone().encode_vec();
+        let hash = cf_guest::digest_with_client_id(client_id, encoded_state.as_slice());
         let mut client = store.private.client_mut(client_id, false)?;
         let state = storage::ClientConsensusState::new(
             processed_time,
             processed_height,
             &state,
         )?;
-        let hash = state.digest(client_id)?;
         client.consensus_states.insert(height, state);
+
 
         let trie_key =
             trie_ids::TrieKey::for_consensus_state(client.index, height);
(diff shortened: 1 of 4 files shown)


solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
```diff
@@ -89,15 +93,17 @@ impl IbcStorage<'_, '_> {
             (head.timestamp_ns, head.block_height)
         };
 
+        let encoded_state = state.clone().encode_vec();
```
mina86, line 96: How about changing `ConsensusState::encode_vec` to take `&self` argument so that we don’t need to `clone` here?

solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
```diff
@@ -237,14 +244,16 @@ impl ibc::ExecutionContext for IbcStorage<'_, '_> {
...
         let mut store = self.borrow_mut();
-        let digest = store
+        let encoded_channel_end = channel_end.clone().encode_vec();
+        let hash = CryptoHash::digest(encoded_channel_end.as_slice());
+        store
             .private
             .port_channel
             .entry(port_channel)
             .or_insert_with(Default::default)
             .set_channel_end(&channel_end)
```
mina86, line 254: Should also update `PortChannelStore::set_channel_end` so that it doesn’t compute and return the hash since that’s no longer needed.

solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
```diff
@@ -135,7 +141,8 @@ impl ibc::ExecutionContext for IbcStorage<'_, '_> {
         msg!("store_connection({}, {:?})", path, connection_end);
         let connection = trie_ids::ConnectionIdx::try_from(&path.0)?;
         let serialised = storage::Serialised::new(&connection_end)?;
-        let hash = serialised.digest();
```
mina86, line 138: We might want to get rid of `Serialised::digest` and `Serialised::digest_with_client` methods since they are now unused and potentially confusing.

mina86 (approved):
I wonder if we should look into storing protobufs instead of Borsh serialised objects.

Protobuf is technically nondeterministic so it’s theoretically possible that different pieces of code generate different representation of the same object. This would lead to invalid proofs. I don’t think this is actually an immediate problem but maybe it’s worth considering?

Protobuf has also better story for backwards compatibility.

On the flip side, Protobuf is potentially slower.

solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
dhruvja: sounds good

solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
dhruvja: gotcha

dhruvja:
in what cases is protobuf non deterministic because if its non deterministic then the values computed by IBC shouldnt match right?

mina86:
Protocol Buffer standard doesn’t mandate order in which fields are encoded.  As such, different encoders may choose to encode the same object in different ways.

In practice this probably doesn’t happen since there is a natural obvious order to put fields in and encoders probably will just stick to that.  In particular, a single implementation will usually generate the same output.

Furthermore, I think Cosmos actually mandates deterministic encoding so even in theory this shouldn’t be an issue.

To be honest I’m not entirely knowledgable in state of protobuf in blockchain.

dhruvja:
I see your point, but what i think is since the ibc uses protobuf for codec of values to which the proofs are proved against, they might have some standard defined. So i think storing values as protobuf wont hurt.

mina86:
SG.

solana/solana-ibc/programs/solana-ibc/src/execution_context.rs
dhruvja: `encode_vec` is defined in `ConsensusState` trait. But we can just add a method for `AnyConsensusState` which encodes by taking a ref.

dhruvja:
I am going to merge this and open another PR for storing protobuf encoded data in the storage.

PR #256, opened 2024-03-20
cf-guest: add method to create new guest client state

dhruvja:
Add a method to create a new guest `ClientState` since it has `PhantomData` as private field. It would be used by the relayer generally to form new clientState

### DIFF
--- a/common/cf-guest/src/client.rs
+++ b/common/cf-guest/src/client.rs
@@ -30,6 +30,22 @@ pub struct ClientState<PK> {
 }
 
 impl<PK: guestchain::PubKey> ClientState<PK> {
+    pub fn new(
+		genesis_hash: CryptoHash,
+		latest_height: guestchain::BlockHeight,
+		trusting_period_ns: u64,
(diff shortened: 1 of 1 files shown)


common/cf-guest/src/client.rs
```diff
@@ -30,6 +30,22 @@ pub struct ClientState<PK> {
...
+		epoch_commitment: CryptoHash,
+		is_frozen: bool,
+	) -> Self {
+		Self {
+			genesis_hash,
+			latest_height,
+			trusting_period_ns,
+			epoch_commitment,
+			is_frozen,
+			_ph: core::marker::PhantomData::<PK>,
```
mina86, line 46:
```suggestion
			_ph: core::marker::PhantomData,
```
should just work, no?

mina86 (approved):

common/cf-guest/src/client.rs
dhruvja: well yes

PR #257, opened 2024-03-20
solana-ibc: store previous host consensus states

dhruvja:
Storing previous consensus states upto a maximum of 20 where the host consensus state can be queried at a particular height. We are storing 20 for the time being but we can go upto 40 or 50 max so that we dont consume a lot of compute units since for `ConnectionOpenAck` takes almost 1.3M CU out of 1.4M.

We need this because the proof height may not always be same as the current height.

### DIFF
--- a/common/guestchain/src/manager.rs
+++ b/common/guestchain/src/manager.rs
@@ -115,12 +115,8 @@ pub enum AddSignatureEffect {
 }
 
 impl AddSignatureEffect {
-    pub fn got_new_signature(self) -> bool {
-        self != Self::Duplicate
-    }
-    pub fn got_quorum(self) -> bool {
-        self == Self::GotQuorum
-    }
+    pub fn got_new_signature(self) -> bool { self != Self::Duplicate }
+    pub fn got_quorum(self) -> bool { self == Self::GotQuorum }
 }
 
 impl<PK: crate::PubKey> ChainManager<PK> {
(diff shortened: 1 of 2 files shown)


mina86:
Who do we need this in ChainManager? ClientConsensusState in storage.rs already has capacity to store multiple states.

dhruvja:
ClientConsensusState stores the consensus state of the counterparty but we need to store the cs of host

dhruvja:
it is called here https://github.com/cosmos/ibc-rs/blob/d773a23aea09bae6cccb7c185db2c4101c090097/ibc-core/ics03-connection/src/handler/conn_open_ack.rs#L116 to verify the proof against it.

common/guestchain/src/manager.rs
```diff
@@ -198,11 +213,19 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
         });
         self.candidates.clear_changed_flag();
+        if self.consensus_states.len() == MAX_CONSENSUS_STATES {
+            self.consensus_states.pop_front();
+        }
+        self.consensus_states.push_back(LocalConsensusState {
+            blockhash: next_block.header.calc_hash().to_vec(),
+            height: next_block.block_height,
+            timestamp: next_block.timestamp_ns,
+        });
```
mina86, line 228: If you swap the operations around you don’t need to clone next_block:

```suggestion
        if self.consensus_states.len() == MAX_CONSENSUS_STATES {
            self.consensus_states.pop_front();
        }
        self.consensus_states.push_back(LocalConsensusState {
            blockhash: next_block.header.calc_hash().to_vec(),
            height: next_block.block_height,
            timestamp: next_block.timestamp_ns,
        });
        self.pending_block = Some(PendingBlock {
            fingerprint,
            next_block,
            signers: Set::new(),
            signing_stake: 0,
        });
        self.candidates.clear_changed_flag();
```

mina86 (approved):

common/guestchain/src/manager.rs
dhruvja: makes sense.

PR #284, opened 2024-04-12
sigverify: sort entries to speed up lookup

mina86:
Change sigverify program so that whenever new entries are add, they are sorted.  While this makes adding signatures more expensive, this helps when checking signatures since a logarithmic search can be used rather than linear scan of the account.


blasrodri (approved):

dhruvja (approved):
LGTM

PR #285, opened 2024-04-16
validator: Improve validator client

dhruvja:
The new validator client has the following improvements.
- not use spinner to send the transaction since it takes forever.
- use program id from config for deriving PDA and calling the program.
- check if pending block exists else create a new block.
- sign only if the current block is unsigned.
- add priority fees as config to let users choose their custom fee.
- submit tx to jito client instead of rpc.
- remove custom accounts structure for staking.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,9 +72,9 @@ serde = "1"
 serde_json = "1"
 serde_bytes = "0.11.14"
 sha2 = { version = "0.10.7", default-features = false }
-solana-client ="1.17.30"
-solana-program ="1.17.30"
-solana-sdk ="1.17.30"
+solana-client = "1.17.30"
+solana-program = "1.17.30"
+solana-sdk = "1.17.30"
 spl-associated-token-account = "2.2.0"
 spl-token = "4.0.0"
 strum = { version = "0.25.0", default-features = false, features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ anchor-spl = "0.29.0"
 ascii = "1.1.0"
 bs58 = { version = "0.5.0", features = ["alloc"] }
 base64 = { version = "0.21", default-features = false, features = ["alloc"] }
+bincode = { version = "1.3.3" }
 borsh = { version = "0.10.3", default-features = false }
 bytemuck = { version = "1.14", default-features = false }
 clap = { version = "4.4.18", features = ["derive"] }
@@ -66,6 +67,7 @@ primitive-types = "0.12.2"
 prost = { version = "0.12.3", default-features = false }
 prost-build = { version = "0.12.3", default-features = false }
 rand = { version = "0.8.5" }
+reqwest = "0.12.3"
 serde = "1"
 serde_json = "1"
 serde_bytes = "0.11.14"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -70,9 +70,9 @@ serde = "1"
 serde_json = "1"
 serde_bytes = "0.11.14"
 sha2 = { version = "0.10.7", default-features = false }
-solana-client = "1.17.7"
-solana-program = "1.17.7"
-solana-sdk = "1.17.7"
+solana-client ="1.17.30"
+solana-program ="1.17.30"
+solana-sdk ="1.17.30"
```
mina86, line 75:
```suggestion
solana-client = "1.17.30"
solana-program = "1.17.30"
solana-sdk = "1.17.30"
```

validator/src/utils.rs
```diff
@@ -125,3 +124,52 @@ pub fn submit_call(
     log::error!("Max retries for signing the block exceeded");
     tx
 }
+
+pub fn submit_generate_block_call(
+    program: &Program<Rc<Keypair>>,
+    validator: &Rc<Keypair>,
+    chain: Pubkey,
+    trie: Pubkey,
+    max_retries: u8,
```
mina86, line 133:
```suggestion
    max_retries: usize,
```

validator/src/utils.rs
```diff
@@ -125,3 +124,52 @@ pub fn submit_call(
...
+    priority_fees: &u64,
+) -> Result<Signature, ClientError> {
+    let mut tries = 0;
+    let mut tx = Ok(Signature::new_unique());
+    while tries < max_retries {
```
mina86, line 138:
```suggestion
    for _ in 0..max_retries {
```

mina86 (approved):

validator/src/utils.rs
```diff
@@ -125,3 +124,52 @@ pub fn submit_call(
...
+        let mut status = true;
```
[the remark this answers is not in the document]
mina86: Sure, but what I mean is that the `status` flag shouldn’t be necessary. Whether the transaction suceeded can be just checked via `tx.is_ok()`.

validator/src/validator.rs
```diff
@@ -23,104 +20,92 @@ pub fn run_validator(config: Config) {
...
+        let chain_account: ChainData = program.account(chain).unwrap();
+        if chain_account.pending_block().unwrap().is_some() {
+            if chain_account
+                .pending_block()
+                .unwrap()
+                .unwrap()
+                .signers
+                .get(&validator.pubkey().into())
+                .is_none()
+            {
```
mina86, line 54: This will work better with condition flipped:

```suggestion
            if pending_block.signers.get(&validator.pubkey().into()).is_some() {
                log::info!("You have already signed the pending block");
                continue;
            }
```

and then the rest of the code doesn’t have to be indentend and it
makes control flow easier to read.

mina86 (approved):

validator/src/utils.rs
```diff
@@ -125,3 +167,56 @@ pub fn submit_call(
...
+        //     e
+        // });
+
+        if let Err(e) = tx.clone() {
+            if !matches!(e, ClientError::SolanaClientError(_)) {
+                return Err(e);
+            }
+        } else if let Ok(tx) = tx {
+            return Ok(tx);
+        }
```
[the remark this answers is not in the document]
dhruvja: makes sense.

PR #290, opened 2024-04-19
Add Dockerfile with build script

mllnd:


docker/build_docker.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+image="composable/emulated-light-client"
+
+docker build \
+    -f ./docker/Dockerfile \
+    -t "${image}:latest" .
```
blasrodri, line 8: can we have commit hash instead?

blasrodri: or tag name if available

mllnd:
Build script adjustments made.
cc @blasrodri

docker/build_docker.sh
```diff
@@ -2,7 +2,8 @@
 set -euo pipefail
 
 image="composable/emulated-light-client"
+tag="$(git describe --tags --exact-match HEAD 2>/dev/null || git rev-parse --short HEAD)"
```
dhruvja, line 5: I think it should refer to HEAD in master branch tbh.

mllnd: @dhruvja I don't 100% agree with that. Ideally, you want to build the image in the same branch which is currently active. At least, that's how we ship & build images in our CI/CDs and local envs.

dhruvja: but there can be some changes in the PRs which can break the validator. So i dont recommend to build from the latest branch.

mikroskeem: It should be validator's responsibility to check out a proper revision

dhruvja: we can merge if you point to the HEAD in master branch.

dhruvja: because for example, when u point to head -> it would install from this https://github.com/ComposableFi/emulated-light-client/pull/334 which would break since this is not deployed yet.

mikroskeem: No, `HEAD` always points to what user has currently checked out from the git repository. Simply put, if user locally checks out `master`, then `HEAD` will point to that; similarly if user checks out #334 branch, `HEAD` will point to that locally.

See [how HEAD works](https://git-scm.com/book/en/v2/Git-Internals-Git-References#ref_the_ref)

dhruvja: I see. I was mistaken then. Looks good to me then.

dhruvja (approved):

PR #296, opened 2024-04-23
restaking: add checking for the Instructions sysvar

mina86:
The Instructions sysvar isn’t used directly by the restaking contract and instead is passed via CPI to the solana-ibc contract.  As such, it’s safe to not verify the account since solana-ibc will perform the verification.  Nonetheless, to introduce another layer of protection, check that the passed account is what we expect.

Note: In addition, remove instruction account from WithdrawalRequest call since it was never used.

### DIFF
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -96,47 +96,51 @@ pub mod restaking {
         // Mint receipt tokens
         token::mint_nft(ctx.accounts.into())?;
 
-        // Call Guest chain program to update the stake if the chain is initialized
-        if guest_chain_program_id.is_some() {
-            let validator_key = match service {
-                Service::GuestChain { validator } => validator,
-            };
-            let borrowed_chain_data =
-                ctx.remaining_accounts[0].data.try_borrow().unwrap();
-            let mut chain_data: &[u8] = &borrowed_chain_data;
-            let chain =
-                solana_ibc::chain::ChainData::try_deserialize(&mut chain_data)
-                    .unwrap();
-            let validator = chain
-                .validator(validator_key)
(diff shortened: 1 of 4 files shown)


solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -96,47 +96,51 @@ pub mod restaking {
...
-            let cpi_program = ctx.remaining_accounts[2].clone();
-            let cpi_ctx =
-                CpiContext::new_with_signer(cpi_program, cpi_accounts, seeds);
-            solana_ibc::cpi::set_stake(cpi_ctx, validator_key, amount)?;
-        }
+        // Call Guest chain program to update the stake if the chain is
+        // initialised.
+        let guest_chain_program_id = match guest_chain_program_id {
+            Some(id) => id,
+            None => return Ok(()),
```
blasrodri, line 103: why is it `Ok`?

mina86: Hmm, it is what existing code does. @dhruvja, should we fail instead? It does look like we’re not recording the stake and just forget about the deposit when guest chain program is not set?

dhruvja: so the `guest_chain_program_id` being `None` indicates that the vaults are only for depositing and not ready for delegating or withdrawing. It was done because we still didnt have the solana-ibc ready when we deployed restaking program. Now that it is set, we can prolly panic but i dont think it affects anyhow.

dhruvja (approved):

solana/restaking/programs/restaking/src/lib.rs
mina86: SG.  If it doesn’t affect correctness, let’s keep it as is and we can change it in another PR if we so desire.

PR #297, opened 2024-04-23
Address various audit findings

mina86:
Each finding is a small change so sending all this as a single PR.  For review I recommend looking at each individual commit rather than full PR. I’m going to marge it as a rebase so this will show up nicely in history as well.


mina86:
Is there something I’m missing or can we get rid of those?

dhruvja:
Ya i think we can get rid of it. it only needed for minting or burning.

mina86:
Added removal of associated_token_program in one place where the code compiled when I did it.

dhruvja (approved):

PR #293, opened 2024-04-21
solana-ibc: fee processing

blasrodri:
**Charging fees**
On send_transfer:
- At the beginning of the function, take the fee

**Refunding fees**
On timeout:
- if the timeout is properly processed upon execution -> refund
On acknowledgement
- if there was an error on the counterparty chain -> refund fees

PR #302, opened 2024-04-25
solana-ibc: Fix decimals

dhruvja:
Different tokens have different decimals so for every new token that is bridged should have the token previously be created by calling `init_mint` which takes original decimals, effective decimals ( decimals if the OG decimals is very large ), token name, token symbol, token uri etc. Right now it can only be created by authority which is the fee collector.

Since solana only supports u64 tokens, we cannot have large decimals since that would overflow u64. So when we mint a new token on solana, we set a fixed decimal for it and if the decimal is less than the original one we chop the rest of the value. This would reduce the precision but thats what we can do for now. 

The decimal on solana would always be equal or less than the decimal of the token on the other chain.

### DIFF
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -1,30 +1,28 @@
+[toolchain]
+
 [features]
 seeds = false
 skip-lint = false
 
 [programs.devnet]
 restaking = "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3"
-solana_ibc = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
+solana_ibc = "9FeHRJLHJSEw4dYZrABHWTRKruFjxDmkLtPmhM5WFYL7"
 
 [programs.localnet]
 restaking = "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3"
-solana_ibc = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
+solana_ibc = "9FeHRJLHJSEw4dYZrABHWTRKruFjxDmkLtPmhM5WFYL7"
 
 [registry]
 url = "https://api.apr.dev"
 
 [provider]
-cluster = "localnet"
-wallet = "~/.config/solana/id.json"
+cluster = "Localnet"
+wallet = "/Users/dhruvjain/.config/solana/id.json"
 
 [workspace]
-members = [
-  "solana/restaking/programs/restaking",
-  "solana/solana-ibc/programs/solana-ibc"
-]
+members = ["solana/restaking/programs/restaking", "solana/solana-ibc/programs/solana-ibc"]
 
 [scripts]
-# test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 solana/restaking/tests/**/*.ts"
 test = "./solana-test.sh"
 
 [test]
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -1,5 +1,3 @@
-[toolchain]
-
 [features]
 seeds = false
 skip-lint = false
@@ -16,13 +14,17 @@ solana_ibc = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
 url = "https://api.apr.dev"
 
 [provider]
-cluster = "Localnet"
+cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
 [workspace]
-members = ["solana/restaking/programs/restaking", "solana/solana-ibc/programs/solana-ibc"]
+members = [
+  "solana/restaking/programs/restaking",
+  "solana/solana-ibc/programs/solana-ibc"
+]
 
 [scripts]
+# test = "yarn run ts-mocha -p ./tsconfig.json -t 1000000 solana/restaking/tests/**/*.ts"
 test = "./solana-test.sh"
 
 [test]
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -6,18 +6,18 @@ skip-lint = false
 
 [programs.devnet]
 restaking = "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3"
-solana_ibc = "9FeHRJLHJSEw4dYZrABHWTRKruFjxDmkLtPmhM5WFYL7"
+solana_ibc = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
 
 [programs.localnet]
 restaking = "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3"
-solana_ibc = "9FeHRJLHJSEw4dYZrABHWTRKruFjxDmkLtPmhM5WFYL7"
+solana_ibc = "9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX"
 
 [registry]
 url = "https://api.apr.dev"
 
 [provider]
 cluster = "Localnet"
-wallet = "/Users/dhruvjain/.config/solana/id.json"
+wallet = "~/.config/solana/id.json"
 
 [workspace]
 members = ["solana/restaking/programs/restaking", "solana/solana-ibc/programs/solana-ibc"]
--- a/solana/solana-ibc/programs/solana-ibc/Cargo.toml
+++ b/solana/solana-ibc/programs/solana-ibc/Cargo.toml
@@ -20,7 +20,7 @@ no-log-ix-name = []
 
 [dependencies]
 anchor-lang.workspace = true
-anchor-spl = {workspace = true, features = ["metadata"] }
+anchor-spl = { workspace = true, features = ["metadata"] }
 base64.workspace = true
 bytemuck = { workspace = true, features = ["must_cast", "zeroable_atomics"] }
 derive_more.workspace = true
--- a/solana/solana-ibc/programs/solana-ibc/Cargo.toml
+++ b/solana/solana-ibc/programs/solana-ibc/Cargo.toml
@@ -20,7 +20,7 @@ no-log-ix-name = []
 
 [dependencies]
 anchor-lang.workspace = true
-anchor-spl.workspace = true
+anchor-spl = {workspace = true, features = ["metadata"] }
 base64.workspace = true
 bytemuck = { workspace = true, features = ["must_cast", "zeroable_atomics"] }
 derive_more.workspace = true
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -59,8 +59,13 @@ pub enum Error {
     /// below required minimum.
     NotEnoughValidators,
 
-    // CPI call from an unidentified program
+    /// CPI call from an unidentified program
     InvalidCPICall,
+
+    /// Unexpected Fee Collector
+    InvalidFeeCollector,
+
+    FeeCollectorChangeProposalNotSet,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -78,6 +78,12 @@ pub enum Error {
     /// When port id, channel id or hased denom passed as arguments
     /// dont match the ones in the packet.
     InvalidSendTransferParams,
+
+    /// Fees can be collected only after a minimum amount
+    InsufficientFeesToCollect,
+
+    /// If both timeout timestamp and timeout height are zero
+    InvalidTimeout,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -74,6 +74,10 @@ pub enum Error {
 
     /// Effective deciamls can either be less than equal to the original decimals but not more.
     InvalidDecimals,
+
+    /// When port id, channel id or hased denom passed as arguments
+    /// dont match the ones in the packet.
+    InvalidSendTransferParams,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -65,8 +65,11 @@ pub enum Error {
     /// Unexpected Fee Collector
     InvalidFeeCollector,
 
+    /// When the new fee collector calls the approve method without the
+    /// new fee collector being set.
     FeeCollectorChangeProposalNotSet,
 
+    /// When an asset is added which already exists
     AssetAlreadyExists,
 
     /// Effective deciamls can either be less than equal to the original decimals but not more.
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -66,6 +66,11 @@ pub enum Error {
     InvalidFeeCollector,
 
     FeeCollectorChangeProposalNotSet,
+
+    AssetAlreadyExists,
+
+    /// Effective deciamls can either be less than equal to the original decimals but not more.
+    InvalidDecimals,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -22,6 +22,8 @@ pub const PACKET_SEED: &[u8] = b"packet";
 pub const SOLANA_IBC_STORAGE_SEED: &[u8] = b"private";
 pub const TRIE_SEED: &[u8] = b"trie";
 pub const MINT_ESCROW_SEED: &[u8] = b"mint_escrow";
+pub const MINT: &[u8] = b"mint";
+pub const ESCROW: &[u8] = b"escrow";
 
 declare_id!("9fd7GDygnAmHhXDVWgzsfR6kSRvwkxVnsY8SaSpSH4SX");
 
@@ -484,7 +486,9 @@ pub struct InitMint<'info> {
               bump, space = 100)]
     mint_authority: UncheckedAccount<'info>,
 
-    #[account(init_if_needed, payer = sender, seeds = [hashed_base_denom.as_ref()],
+    #[account(init_if_needed, payer = sender,
+              seeds = [MINT, port_id.as_bytes(), channel_id_on_b.as_bytes(),
+                       hashed_base_denom.as_ref()],
               bump, mint::decimals = 6, mint::authority = mint_authority)]
     token_mint: Account<'info, Mint>,
 
@@ -608,7 +612,7 @@ pub struct SendTransfer<'info> {
     #[account(mut)]
     token_mint: Option<Box<Account<'info, Mint>>>,
     #[account(init_if_needed, payer = sender, seeds = [
-        port_id.as_bytes(), channel_id.as_bytes(), hashed_base_denom.as_ref()
+        ESCROW, port_id.as_bytes(), channel_id.as_bytes(), hashed_base_denom.as_ref()
     ], bump, token::mint = token_mint, token::authority = mint_authority)]
     escrow_account: Option<Box<Account<'info, TokenAccount>>>,
     #[account(mut)]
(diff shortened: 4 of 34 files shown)


Anchor.toml
```diff
@@ -1,30 +1,28 @@
...
-wallet = "~/.config/solana/id.json"
+cluster = "Localnet"
+wallet = "/Users/dhruvjain/.config/solana/id.json"
 
 [workspace]
-members = [
-  "solana/restaking/programs/restaking",
-  "solana/solana-ibc/programs/solana-ibc"
-]
+members = ["solana/restaking/programs/restaking", "solana/solana-ibc/programs/solana-ibc"]
```
mina86, line 23: Those changes are probably unnecessary.

solana/solana-ibc/programs/solana-ibc/Cargo.toml
```diff
@@ -20,7 +20,7 @@ no-log-ix-name = []
 
 [dependencies]
 anchor-lang.workspace = true
-anchor-spl.workspace = true
+anchor-spl = {workspace = true, features = ["metadata"] }
```
mina86, line 23:
```suggestion
anchor-spl = { workspace = true, features = ["metadata"] }
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -191,6 +201,68 @@ pub mod solana_ibc {
...
+
+        let signer = ctx.accounts.fee_collector.key();
+
+        if private_storage.fee_collector == Pubkey::default() {
+            private_storage.fee_collector = new_fee_collector;
+        } else if signer == private_storage.fee_collector {
+            private_storage.new_fee_collector_proposal = Some(new_fee_collector)
+        } else {
+            return Err(error::Error::InvalidFeeCollector)
+                .map_err(|err| error!((&err)));
```
mina86, line 221:
```suggestion
            return Err(error!(error::Error::InvalidFeeCollector));
```

should just work.  Here and below.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -191,6 +201,68 @@ pub mod solana_ibc {
...
+            }
+            private_storage.fee_collector = new_admin;
+            private_storage.new_fee_collector_proposal = None;
+        } else {
+            return Err(error::Error::FeeCollectorChangeProposalNotSet)
+                .map_err(|err| error!((&err)));
+        }
+
+        Ok(())
+    }
```
mina86, line 247: I think we should just require a transaction signed by both. It makes things easier on the contract side. And on the client side it would effectively work as multisig. Does that make sense?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -191,6 +201,68 @@ pub mod solana_ibc {
...
+
+        let total_fees_collected = fee_collector_balance - minimum_rent_balance;
+
+        **ctx.accounts.fee_account.try_borrow_mut_lamports().unwrap() -=
+            total_fees_collected;
+        **ctx.accounts.fee_collector.try_borrow_mut_lamports().unwrap() +=
+            total_fees_collected;
+
+        Ok(())
+    }
```
mina86, line 263: I don’t understand what this is supposed to do. Why are we transferring all balance from an account? And why are we assuming the account has zero size? (I’m also not sure if we don’t need to use system_instruction::transfer to do the transfer).

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: but its 2 different accounts signing on those methods. The first one is signed by existing one and the second is signed by the new fee collector.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: the fees are collected in a PDA so that the fees can be returned during timeouts and failed transfers. So this method transfers the fee collected back to the fee collector account set.

dhruvja: But then i think we should transfer only part of the fees

solana/solana-ibc/programs/solana-ibc/src/lib.rs
mina86: Yes, but transaction can be signed by multiple accounts so it should be possible. The difference is that the accounts holders would need to coordinate off-chain to execute the transaction.

dhruvja: i think it would be a bit convoluted though

dhruvja: for the users.

mina86: Do we change fee collector often? I’d rather move complexity to client than having it in the contract. Solana has interface for partial signing: https://solana-labs.github.io/solana-web3.js/classes/Transaction.html#partialSign.partialSign-1

dhruvja: well ideally not

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -476,25 +601,72 @@ pub struct ChainWithVerifier<'info> {
...
+}
+
+#[derive(Accounts)]
+pub struct CollectFees<'info> {
     #[account(mut)]
+    fee_collector: Signer<'info>,
+
+    /// The account holding private IBC storage.
+    #[account(mut, seeds = [SOLANA_IBC_STORAGE_SEED],
+    bump, has_one = fee_collector)]
```
mina86, line 621: Uh? What’s the purpose of has_one here?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: i think partial sign could work

dhruvja: just worried how we would do with multisig

solana/solana-ibc/programs/solana-ibc/src/lib.rs
mina86: I don’t think it should be that complicated. ‘Multisig’ is perhaps the wrong term to use her, but from what I’m seeing it should be relatively easy to prepare a transaction and than have it signed by multiple accounts via the partialSign or even addSignature. There is coordination cost on the client side but the two accounts have to coordinate anyway in some capacity.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: only fee collector can call this method.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: well it kinda makes sense.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: Fixed in #293

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -259,8 +259,23 @@ pub struct PrivateStorage {
     pub port_channel: map::Map<trie_ids::PortChannelPK, PortChannelStore>,
 
     pub channel_counter: u32,
+
+    pub fee_collector: Pubkey,
+
+    pub new_fee_collector_proposal: Option<Pubkey>,
+
+    pub assets: Vec<Asset>,
```
vmarkushin, line 267: Why not to use `Map` from denom?

vmarkushin (commented):
The main logic LGTM

solana/solana-ibc/programs/solana-ibc/src/error.rs
```diff
@@ -59,8 +59,18 @@ pub enum Error {
...
-    // CPI call from an unidentified program
+    /// CPI call from an unidentified program
     InvalidCPICall,
+
+    /// Unexpected Fee Collector
+    InvalidFeeCollector,
+
+    FeeCollectorChangeProposalNotSet,
+
+    AssetAlreadyExists,
```
mina86, line 70: Could you add docstrings to those as well?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -25,6 +26,10 @@ pub const MINT_ESCROW_SEED: &[u8] = b"mint_escrow";
 pub const MINT: &[u8] = b"mint";
 pub const ESCROW: &[u8] = b"escrow";
 
+pub const FEE_SEED: &[u8] = b"fee";
+
+pub const FEE_AMOUNT_IN_LAMPORTS: u64 = 10_000_000; // 0.01 SOL
```
mina86, line 31:
```suggestion
pub const FEE_AMOUNT_IN_LAMPORTS: u64 =
    solana_program::native_token::LAMPORTS_PER_SOL / 100;
```

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -423,7 +487,7 @@ fn anchor_test_deliver() -> Result<()> {
         &associated_token_addr,
         &authority.pubkey(),
         &[&authority.pubkey()],
-        1000000000,
+        1000000000_000_000_000,
```
mina86, line 490: We have constant for that?

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -20,19 +20,27 @@ use anchor_spl::associated_token::get_associated_token_address;
...
 pub const STAKING_PROGRAM_ID: &str =
     "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3";
 pub const WRITE_ACCOUNT_SEED: &[u8] = b"write";
+pub const TOKEN_NAME: &str = "RETARDIO";
+pub const TOKEN_SYMBOL: &str = "RTRD";
+pub const TOKEN_URI: &str = "https://github.com";
 // const BASE_DENOM: &str = "PICA";
 
-const TRANSFER_AMOUNT: u64 = 1000000;
+const TRANSFER_AMOUNT: u64 = 1000000_000_000_000;
```
mina86, line 40:
```suggestion
const TRANSFER_AMOUNT: u64 = 1_000_000_000_000_000;
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -112,11 +114,42 @@ impl TokenTransferExecutionContext for IbcStorage<'_, '_> {
...
+                    &amt.denom.base_denom.as_str().as_bytes(),
+                ))
+            })
+            .ok_or(TokenTransferError::InvalidToken)?;
+
+        let converted_amount = convert_decimals(
+            &amt.amount,
+            asset.original_decimals,
+            asset.effective_decimals_on_sol,
+        )
```
mina86, line 135: I’m a bit confused.  We have conversion from original decimals to decimals on Solana, but where does conversion in the opposite direction happen? When someone transfers tokens back the conversion must happen in opposite and I don’t see it in the code.

solana/solana-ibc/programs/solana-ibc/src/error.rs
dhruvja: sure

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: well the thing is the amount in the packet always contain with original decimals, it only during mint and burning that we chop them off.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -201,10 +273,78 @@ pub mod solana_ibc {
...
+        original_decimals: u8,
+        token_name: String,
+        token_symbol: String,
+        token_uri: String,
     ) -> Result<()> {
+        let private_storage = &mut ctx.accounts.storage;
+
+        if effective_decimals > original_decimals {
+            return Err(error!(error::Error::InvalidDecimals));
+        }
```
mina86, line 289: With convert_decimals being able to handle conversion in whatever direction do we need this check?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: closing this as we decided that it can get complicated

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: well we wont be having a lot of assets so i dont think map would do us any favor

mina86: The `Map` is a linear map and having it will save you on having to do the lookup iteration.  It’s the same type we’re using for consensus states.

dhruvja: hmm alright, i can use that

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -73,6 +82,25 @@ fn get_token_mint(denom: &PrefixedDenom) -> Result<Pubkey, TokenTransferError> {
...
+    match (dest_port_id, dest_channel_id) {
+        (Some(port_id), Some(channel_id)) => prefixed_denom.remove_trace_prefix(&TracePrefix::new(
+            PortId::from_str(port_id).unwrap(),
+            ChannelId::from_str(channel_id).unwrap(),
+        )),
+        (_, _) => ()
+        };
+    let full_denom = prefixed_denom.to_string();
+    CryptoHash::digest(full_denom.as_bytes())
+}
```
mina86: I’d like to see some unit tests for this because it becomes unclear
what this actually dose.  From what I understand, it takes prefix such
as `foo/bar/baz/qux` and then returns hash of `baz/qux`?  Except if
prefix has only one component in which case it returns hashs of the
full prefix.

Also if that’s what this function is doing than using `match_indices`
may be better:

```suggestion
/// Removes the destination source and port id and
/// returns the hash of full denom.
pub fn get_hashed_full_denom(denom: &PrefixedDenom) -> CryptoHash {
    let trace_path = denom.trace_path.to_string();
    let idx = trace_path.match_indices('/').nth(1).map_or(0, |idx| idx + 1);
    CryptoHash::digest(trace_path.as_bytes()[idx..])
}
```

mina86: Also, shouldn’t this take base denom into account?  Something like

    CryptoHash::digestv(&[trace_path.as_bytes()[idx..], denom.base_denom.as_str().as_bytes()])

dhruvja: if you convert the `PrefixedDenom` to string then no need but if you only consider `trace_path` then you need to do it. So as per your suggestion we need to append base denom too.

dhruvja: tbh, this is only called during minting and burning in which it should always have a prefix

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -232,11 +238,10 @@ pub mod solana_ibc {
...
             if signer != new_admin {
                 return Err(error!(error::Error::InvalidFeeCollector));
+            } else {
+                return Err(error::Error::FeeCollectorChangeProposalNotSet)
+                    .map_err(|err| error!((&err)));
             }
-            private_storage.fee_collector = new_admin;
-            private_storage.new_fee_collector_proposal = None;
-        } else {
-            return Err(error!(error::Error::FeeCollectorChangeProposalNotSet));
```
mina86, line 239: Bad merge?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -767,7 +863,7 @@ pub struct SendTransfer<'info> {
     #[account(mut)]
     token_mint: Option<Box<Account<'info, Mint>>>,
     #[account(init_if_needed, payer = sender, seeds = [
-        ESCROW, port_id.as_bytes(), channel_id.as_bytes(), hashed_base_denom.as_ref()
+        ESCROW, port_id.as_bytes(), channel_id.as_bytes(), hashed_full_denom.as_ref()
```
mina86, line 866: If we’re hashing full denome do we still need port and channel in the seeds?

solana/solana-ibc/programs/solana-ibc/src/storage.rs
```diff
@@ -308,8 +308,17 @@ pub struct PrivateStorage {
     pub fee_collector: Pubkey,
 
     pub new_fee_collector_proposal: Option<Pubkey>,
+
+    pub assets: map::Map<CryptoHash, Asset>,
+}
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone, Debug)]
```
mina86, line 315: For other types we used borsh instead of the alias so let’s do that I guess?

```suggestion
#[derive(Clone, Debug, borsh::BorshSerialize, borsh::BorshDeserialize)]
```

solana/solana-ibc/programs/solana-ibc/src/transfer/mod.rs
```diff
@@ -9,7 +9,7 @@ use crate::ibc::apps::transfer::types::packet::PacketData;
 use crate::ibc::apps::transfer::types::proto::transfer::v2::FungibleTokenPacketData;
 use crate::storage::IbcStorage;
 
-mod impls;
+pub mod impls;
```
mina86, line 12:
```suggestion
pub(crate) mod impls;
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: Yes im pushing a fix for this

solana/solana-ibc/programs/solana-ibc/src/storage.rs
dhruvja: sounds good.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: ya lol

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -626,24 +696,42 @@ pub struct CollectFees<'info> {
...
+#[instruction(decimals: u8, hashed_full_denom: CryptoHash)]
 pub struct InitMint<'info> {
-    #[account(mut)]
+    #[account(mut, constraint = sender.key == &storage.fee_collector)]
     sender: Signer<'info>,
 
+    #[account(
+        mut,
+        seeds = [
+            b"metadata".as_ref(),
```
mina86, line 706: Should we turn it into a `pub const`?

dhruvja: sure

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: closing since we have removed this.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -430,14 +492,23 @@ pub mod solana_ibc {
             .map_err(|err| error!((&err)))
     }
 
-    #[allow(unused_variables)]
+    /// The port_id, channel_id and hashed_full_denom are passed
```
mina86, line 495: port_id and channel_id are no longer passed.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -527,3 +552,104 @@ fn check_amount_overflow(amount: Amount) -> Result<u64, TokenTransferError> {
...
+        ]);
+        // Remove base denom
+        trace_path_split.pop();
+        let trace_path = TracePath::try_from(trace_path_split).unwrap();
+        let expected_trace_path = vec![
+            TracePrefix::new(PortId::transfer(), ChannelId::new(0)),
+            TracePrefix::new(PortId::transfer(), ChannelId::new(1)),
+        ];
+        assert_eq!(trace_path, expected_trace_path.into());
+    }
```
mina86, line 654: Is this still relevant?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: no its not.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: right

PR #311, opened 2024-04-28
Update tendermint dependency

mina86:
Update tendermint dependency to include our sign bytes reuse code.
The fork is mostly Tendermint 0.36.0 with just the sign bytes encoding
optimisation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,10 +120,10 @@ aes-gcm-siv = { git = "https://github.com/RustCrypto/AEADs", rev = "6105d7a5591a
 curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dalek", rev = "8274d5cbb6fc3f38cdc742b4798173895cd2a290" }
 
 # Uses solana sys call as default hashing
-tendermint                       = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
-tendermint-light-client          = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -120,10 +120,10 @@ aes-gcm-siv = { git = "https://github.com/RustCrypto/AEADs", rev = "6105d7a5591a
 curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dalek", rev = "8274d5cbb6fc3f38cdc742b4798173895cd2a290" }
 
 # Uses solana sys call as default hashing
-tendermint                       = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
-tendermint-light-client          = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
-tendermint-light-client-verifier = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
-tendermint-proto                 = { git = "https://github.com/mina86/tendermint-rs", rev = "7cad235b651728dbfe6d2bfbd9110c329f0d6e7e" }
+tendermint                       = { git = "https://github.com/mina86/tendermint-rs", rev = "7b674c3693076de393069ab80bc1f703f639949a" }
```
dhruvja, line 123: This is the latest commit with timestamp replacement during encoding
```suggestion
tendermint                       = { git = "https://github.com/mina86/tendermint-rs", rev = "9f157c06f9053940bd182f4b3e8e958e5731d0c7" }
```

dhruvja (approved):

Cargo.toml
mina86: I assumed we concluded that the reuse of the canonical vote isn’t buying as much and thus isn’t really worth it?

dhruvja: until we have a better solution, lets keep this.

PR #312, opened 2024-04-29
solana-ibc: use wallet addresses instead of token accounts in packets

dhruvja:
Currently we were using token accounts as sender/receiver in the packets. But coming to think of it, its better to use wallet addresses since token accounts can be easily derived from the token mint and wallet address. It also makes it easy to send a transfer from other chain by just sending it to wallet address rather than deriving token account on client side and sending it since that cannot be trusted. Additionally we have checks to ensure that the token account to which the tokens are transferred to are always owned by the sender/receiver.

Also in `unescrow_validate` using `Unescrow` tag instead of `Escrow` tag for validation.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/tests.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/tests.rs
@@ -452,8 +452,8 @@ fn anchor_test_deliver() -> Result<()> {
         port_id.clone(),
         true,
         channel_id_on_a.clone(),
-        associated_token_addr,
-        receiver_token_address,
+        authority.pubkey(),
+        receiver.pubkey(),
     );
 
     let account_balance_before = sol_rpc_client
@@ -519,8 +519,8 @@ fn anchor_test_deliver() -> Result<()> {
         channel_id_on_b.clone(),
         channel_id_on_a.clone(),
         2,
-        sender_token_address,
-        receiver_token_address,
+        authority.pubkey(),
+        receiver.pubkey(),
         String::from("Tx from destination chain"),
     );
     let proof_height_on_a = mock_client_state.header.height;
@@ -587,8 +587,8 @@ fn anchor_test_deliver() -> Result<()> {
         port_id.clone(),
         false,
         channel_id_on_a.clone(),
-        receiver_token_address,
-        sender_token_address,
+        receiver.pubkey(),
+        authority.pubkey(),
     );
 
     let account_balance_before = sol_rpc_client
@@ -656,8 +656,8 @@ fn anchor_test_deliver() -> Result<()> {
         channel_id_on_b.clone(),
         channel_id_on_a.clone(),
         3,
-        sender_token_address,
-        receiver_native_token_address,
+        authority.pubkey(),
+        receiver.pubkey(),
         String::from("Tx from Source chain"),
     );
 
--- a/solana/solana-ibc/programs/solana-ibc/src/tests.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/tests.rs
@@ -16,6 +16,7 @@ use anchor_client::solana_sdk::signature::{
 use anchor_client::solana_sdk::transaction::Transaction;
 use anchor_client::{Client, Cluster};
 use anchor_lang::solana_program::system_instruction::create_account;
+use anchor_lang::AnchorSerialize;
 use anchor_spl::associated_token::get_associated_token_address;
 use anyhow::Result;
 use ibc::apps::transfer::types::msgs::transfer::MsgTransfer;
@@ -26,6 +27,7 @@ use crate::{
     accounts, chain, ibc, instruction, ix_data_account, CryptoHash,
     MINT_ESCROW_SEED,
 };
+
 const IBC_TRIE_PREFIX: &[u8] = b"ibc/";
 pub const STAKING_PROGRAM_ID: &str =
     "8n3FHwYxFgQCQc2FNFkwDUf9mcqupxXcCvgfHbApMLv3";
@@ -88,6 +90,14 @@ fn anchor_test_deliver() -> Result<()> {
             .unwrap()
             .pubkey();
 
+    let fee_collector_keypair = Rc::new(Keypair::new());
+    let fee_collector = fee_collector_keypair.pubkey();
+
+    let sol_rpc_client = program.rpc();
+    let _airdrop_signature =
+        airdrop(&sol_rpc_client, authority.pubkey(), lamports);
+    let _airdrop_signature = airdrop(&sol_rpc_client, fee_collector, lamports);
+
     // Build, sign, and send program instruction
     let storage = Pubkey::find_program_address(
         &[crate::SOLANA_IBC_STORAGE_SEED],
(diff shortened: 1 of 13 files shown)


solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -287,10 +293,31 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
             .ok_or(TokenTransferError::ParseAccountFailure)?;
-        if !account.0.eq(token_account.key) {
+        let receiver = accounts
+            .receiver
+            .as_ref()
+            .ok_or(TokenTransferError::ParseAccountFailure)?;
+
+        let receiver_token_account = get_token_account(account.0, token_mint);
+
+        if !account.0.eq(receiver.key) {
```
mina86, line 303:
```suggestion
        if account.0 != *receiver.key {
```
and same throughout the PR.

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -73,6 +73,12 @@ fn get_token_mint(denom: &PrefixedDenom) -> Result<Pubkey, TokenTransferError> {
     Ok(Pubkey::find_program_address(&seeds, &crate::ID).0)
 }
 
+fn get_token_account(owner: Pubkey, token_mint: Pubkey) -> Pubkey {
```
mina86, line 76:
```suggestion
fn get_token_account(owner: &Pubkey, token_mint: &Pubkey) -> Pubkey {
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -388,17 +436,26 @@ impl IbcStorage<'_, '_> {
...
-        accounts
-            .token_account
-            .as_ref()
-            .filter(|token_account| account.0.eq(token_account.key))
-            .ok_or(TokenTransferError::ParseAccountFailure)?;
-
+        // We only need to check for sender/receiver since the token account
+        // is always derived from the token mint so if sender/receiver are right,
+        // the token account would be right as well.
         let ok = match op {
```
mina86, line 442: I’m not too happy with this match here.  On one side we have `ok_or`
with question mark operator.  On the other the `ok` variable is still
used.  Then again I don’t know if there’s some cleaner way to do it.
Perhaps something like:

```
        match op {
            EscrowOp::Escrow => {
                accounts
                    .sender
                    .as_deref()
                    .filter(|sender| sender.is_signer)
            },
            EscrowOp::Unescrow => {
                accounts
                    .receiver
                    .as_deref()
                    .filter(|_| accounts.mint_authority.is_some())
            },
        }
            .filter(|acc| account.0 == *acc.key)
            .map(|_| ())
            .ok_or(TokenTransferError::ParseAccountFailure)
```

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: gotcha

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: right

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
dhruvja: i think above is good enough

solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -720,9 +720,9 @@ fn anchor_test_deliver() -> Result<()> {
         channel_id_on_b.clone(),
         channel_id_on_a.clone(),
         3,
-        sender_token_address,
-        receiver_native_token_address,
-        String::new(),
+        authority.pubkey(),
+        receiver.pubkey(),
+        String::from(""),
```
mina86, line 725: But why not `String::new`?

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -287,10 +293,31 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+            .as_ref()
+            .ok_or(TokenTransferError::ParseAccountFailure)?;
+
+        let receiver_token_account = get_token_account(&account.0, &token_mint);
+
+        if account.0 != *receiver.key {
+            msg!("Token account not found {} {:?}", account, receiver.key);
             return Err(TokenTransferError::ParseAccountFailure);
         }
         if !token_mint.eq(token_mint_account.key) {
```
mina86, line 307:
```suggestion
        if token_mint != *token_mint_account.key {
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -287,10 +293,31 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+            msg!(
+                "Token mint not found {:?} {:?}",
+                token_mint,
+                token_mint_account.key
+            );
+            return Err(TokenTransferError::ParseAccountFailure);
+        }
+        if !receiver_token_account.eq(token_account.key) {
```
mina86, line 315:
```suggestion
        if receiver_token_account != *token_account.key {
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -328,10 +355,31 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+            .as_ref()
+            .ok_or(TokenTransferError::ParseAccountFailure)?;
+
+        let sender_token_account = get_token_account(&account.0, &token_mint);
+
+        if account.0 != *sender.key {
+            msg!("Token account not found {} {:?}", account, sender.key);
             return Err(TokenTransferError::ParseAccountFailure);
         }
         if !token_mint.eq(token_mint_account.key) {
```
mina86, line 369:
```suggestion
        if token_mint != *token_mint_account.key {
```

solana/solana-ibc/programs/solana-ibc/src/transfer/impls.rs
```diff
@@ -328,10 +355,31 @@ impl TokenTransferValidationContext for IbcStorage<'_, '_> {
...
+            msg!(
+                "Token mint not found {:?} {:?}",
+                token_mint,
+                token_mint_account.key
+            );
+            return Err(TokenTransferError::ParseAccountFailure);
+        }
+        if !sender_token_account.eq(token_account.key) {
```
mina86, line 377:
```suggestion
        if sender_token_account != *token_account.key {
```

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/tests.rs
dhruvja: sure

PR #313, opened 2024-04-29
solana-ibc: limit number of stored consensus states

mina86:


dhruvja:
I think since we are removing the states ourselves, we should prolly return an empty array when requesting for previous consensus states used for pruning.
https://github.com/ComposableFi/emulated-light-client/blob/922ae3f0c54a967ebc9c957f7627142fb160536b/solana/solana-ibc/programs/solana-ibc/src/client_state.rs#L200-L216

dhruvja:
> Yes, SG. I didn’t do it initially thinking to handle transitional period where we have more than the limit of states at which point we would need another method to systematically prune old states. But if we start from a list which is already no more than the limit we don’t need light clients to prune states.

ya exactly.

dhruvja (approved):

PR #314, opened 2024-05-01
solana-ibc: update TS sdk for send transfer

dhruvja:


mina86 (commented):
I mean it looks good, but if we’re deleting code, can we just delete code rather than comment it? ;)

dhruvja:
oh well, it wasnt meant to be deleted.

mina86 (approved):

PR #316, opened 2024-05-03
convert to events and add event for consensus

mina86:
As per comment, perhaps it’s better to use proper event? This PR does three things:
- Converts the behaviour from logging to emitting an event.
- Adds an event for consensus state.  I honestly don’t know how useful that is? If we don’t need it it might be better to skip it.
- Changes to output proto encoding rather then borsh. This is because proto is what is being hashed and stored in the trie, so perhaps that’s more convenient than outputting borsh?
Thoughts?


dhruvja:
I dont think emitting consensus state is any helpful as of now, but i think its better to keep and if we need it in future we can use it.

Regarding encoding with proto instead of borsh, i dont think it really matters much. At the end of the day the only thing that matters is which is more efficient.

mina86:
OK, let’s drop consensus then so we don’t spend CU, memory or log limit on it and use borsh format since it is a bit shorter I believe.

dhruvja (approved):

PR #318, opened 2024-05-05
restaking: record the last received height when collecting rewards in withdrawal request

dhruvja:
When requesting for withdrawal, the pending rewards are transferred to the user but the height at which the rewards were claimed was not updated. This fixes that.


dhruvja:
`local.json` not again!!!

mina86 (approved):

PR #325, opened 2024-05-17
ts-sdk: perform proper serialization of amount to u256

dhruvja:
Use LE format to convert the bigint to 32 byte array instead of 8 byte done previously.

### DIFF
--- a/solana/solana-ibc/tests/utils.ts
+++ b/solana/solana-ibc/tests/utils.ts
@@ -14,11 +14,15 @@ import {
 import bs58 from "bs58";
 import { solanaIbcProgramId } from "./constants";
 
-export function getInt64Bytes(x: number) {
-  let y = Math.floor(x / 2 ** 32);
-  return [y, y << 8, y << 16, y << 24, x, x << 8, x << 16, x << 24].map(
-    (z) => z >>> 24
-  );
+export function numberTo32ByteBuffer(num: bigint): Uint8Array {
+  const buffer = Buffer.alloc(32);
+  let numberHex = num.toString(16);
+  if (numberHex.length % 2 !== 0) {
+    numberHex = "0" + numberHex;
+  }
+  const numberBytes = Buffer.from(numberHex, "hex");
+  numberBytes.reverse().copy(buffer, 0);
+  return new Uint8Array(buffer);
 }
 
 export function hexToBytes(hex: string) {
--- a/solana/solana-ibc/tests/solana-ibc.ts
+++ b/solana/solana-ibc/tests/solana-ibc.ts
@@ -11,7 +11,7 @@ import {
   sendAndConfirmTransaction,
 } from "@solana/web3.js";
 import { solanaIbcProgramId, rpcUrl, depositor } from "./constants";
-import { getInt64Bytes, hexToBytes, getGuestChainAccounts } from "./utils";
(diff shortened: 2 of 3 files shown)


solana/solana-ibc/tests/utils.ts
```diff
@@ -14,11 +14,15 @@ import {
...
-  );
+export function numberTo32ByteBuffer(num: bigint): Uint8Array {
+  const buffer = Buffer.alloc(32);
+  let numberHex = num.toString(16);
+  if (numberHex.length % 2 !== 0) {
+    numberHex = "0" + numberHex;
+  }
+  const numberBytes = Buffer.from(numberHex, "hex");
+  numberBytes.reverse().copy(buffer, 0);
+  return new Uint8Array(buffer);
```
mina86, line 25: Perhaps worth checking range?  Also perhaps this is more straightforward:

```suggestion
	const n = BigInt.asUintN(256, num);
	if (num < 0 || n < num) {
		throw new RangeError();
	}
	const numHex = '00'.repeat(32) + n.toString(16);
	const numBuf = Buffer.from(numHex.substring(numHex.length - 64), 'hex');
	return new Uint8Array(numBuf.reverse());
```

not tested.

dhruvja: ya i think the above code is not right. Since the U256 is [u64; 4] and convert to le, bytes are assigned from the back of the 32 byte array which is prolly not right. It should rather be split into 4 arrays and then converted to LE.
Something like this.
```ts
export function numberTo32ByteBuffer(num: bigint): Uint8Array {
  // Create a buffer of 32 bytes initialized with zeros
  let buffer = Buffer.alloc(32);

  // Convert the BigInt to a hexadecimal string
  let numberHex = num.toString(16);
  if (numberHex.length % 2 !== 0) {
    numberHex = "0" + numberHex; // Ensure the hex string has an even length
  }

  // Create a buffer from the hexadecimal string
  let numberBytes = Buffer.from(numberHex, "hex");

  // Copy the number bytes to the end of the 32-byte buffer
  numberBytes.copy(buffer, 32 - numberBytes.length);
  let uintBuffer = new Uint8Array(buffer);
  // split above array into 4 chunks of 8 bytes each
  let uintBufferChunks = [];
  for (let i = 0; i < uintBuffer.length; i += 8) {
    uintBufferChunks.push(uintBuffer.slice(i, i + 8));
  }
  let final_uintBuffer: Array<number> = [];
  final_uintBuffer = final_uintBuffer.concat(...uintBufferChunks[3]);
  final_uintBuffer = final_uintBuffer.concat(...uintBufferChunks[2]);
  final_uintBuffer = final_uintBuffer.concat(...uintBufferChunks[1]);
  final_uintBuffer = final_uintBuffer.concat(...uintBufferChunks[0]);
  return new Uint8Array(final_uintBuffer);
}
```

mina86: This is still unresolved, right?

dhruvja: well actually no, the above seems to work. The FE is using that only.

mina86: But this will break if number is over u64::MAX, no? Because of the weird way U256 is serialised in IBC (each u64 is serialised using little endian but the four u64 are serialised in big endian order IIRC).

PR #326, opened 2024-05-23
remove license for the repository

dhruvja:
Removing the existing BS license we have currently and making the project completely open source.


mina86:
Probably replace the LICENSE file with a GPL license.  Current license says GPL 2 so that’s https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt. For GPL 3 it would be https://www.gnu.org/licenses/gpl-3.0.txt.  As for change itself that’s something up to Blas I guess.

PR #328, opened 2024-05-24
solana-ibc: update fees

mina86:


mina86:
Inside the account? You mean as a configuration option?

dhruvja:
yes inside an account, so that they can just execute the method instead of pushing an upgrade.

PR #333, opened 2024-05-29
solana-ibc: remove send packet method

dhruvja:
Removing send packet method for now since we dont support cross chain messaging yet.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -412,97 +412,100 @@ pub mod solana_ibc {
         )
     }
 
-    /// Should be called after setting up client, connection and channels.
-    pub fn send_packet<'a, 'info>(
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -412,97 +412,100 @@ pub mod solana_ibc {
...
-        }
-
-        // Since we do all the checks present in validate above, there is no
-        // need to call validate again.  Hence validate is only called during
-        // tests.
-        ::ibc::core::channel::handler::send_packet_execute(&mut store, packet)
-            .map_err(error::Error::ContextError)
-            .map_err(|err| error!((&err)))
-    }
+    /// This method is removed since we dont support cross chain messaging yet
```
mina86, line 415: Let’s just delete it. No need to keep it. We can look into history to get it back if we need to.

dhruvja: ok lets do that then

mina86 (approved):

PR #334, opened 2024-05-29
restaking: Restaking oracle

dhruvja:
Adding support to get price feed from oracle so that we can support staking of assets other than SOL and LSTs. The method gets the token price from oracle and then is divided by sol price to get in terms of sol. If the price feed id doesnt exist for tokens, then the solana price feed id is used as default.

### DIFF
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -33,9 +33,15 @@ upgradeable = true
 
 [test.validator]
 bind_address = "0.0.0.0"
-url = "https://api.devnet.solana.com"
+url = "https://api.mainnet-beta.solana.com"
 ledger = ".anchor/test-ledger"
 rpc_port = 8899
 
 [[test.validator.clone]]
 address = "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s"
+
+[[test.validator.clone]]
+address = "Dpw1EAVrSB1ibxiDQyTAW6Zip3J4Btk2x4SgApQCeFbX"
+
+[[test.validator.clone]]
+address = "7UVimffxr9ow1uXYxsr4LHAcV58mLzhmwaeKvJ1pjLiE"
--- a/solana/restaking/programs/restaking/src/lib.rs
+++ b/solana/restaking/programs/restaking/src/lib.rs
@@ -95,23 +95,39 @@ pub mod restaking {
             .unwrap_or(&sol_price_feed_id);
 
         let final_amount_in_sol = if token_feed_id == &sol_price_feed_id {
-            // let maximum_age_in_sec: u64 = 30;
-            let feed_id: [u8; 32] = get_feed_id_from_hex(token_feed_id)?;
-            let token_price =
-                token_price_update.get_price_unchecked(&feed_id)?;
+            let (token_price, sol_price) = if cfg!(feature = "mocks") {
+                let feed_id: [u8; 32] = get_feed_id_from_hex(token_feed_id)?;
+                let sol_price = sol_price_update.get_price_unchecked(
+                    &get_feed_id_from_hex(SOL_PRICE_FEED_ID)?,
+                )?;
+                let token_price =
+                    token_price_update.get_price_unchecked(&feed_id)?;
+                (token_price, sol_price)
+            } else {
+                let maximum_age_in_sec: u64 = 30;
+                let feed_id: [u8; 32] = get_feed_id_from_hex(token_feed_id)?;
+                let sol_price = sol_price_update.get_price_no_older_than(
+                    &Clock::get()?,
+                    maximum_age_in_sec,
+                    &get_feed_id_from_hex(SOL_PRICE_FEED_ID)?,
+                )?;
+                let token_price = token_price_update.get_price_no_older_than(
+                    &Clock::get()?,
+                    maximum_age_in_sec,
+                    &feed_id,
+                )?;
+                (token_price, sol_price)
+            };
+
             let token_decimals = ctx.accounts.token_mint.decimals;
 
-            let sol_price = sol_price_update.get_price_unchecked(
-                &get_feed_id_from_hex(SOL_PRICE_FEED_ID)?,
-            )?;
+            let amount_in_sol_decimals = (amount *
+                10u64.pow(SOL_DECIMALS as u32)) /
+                10u64.pow(token_decimals as u32);
 
-            let amount_in_sol_decimals = (amount
-                * 10u64.pow(SOL_DECIMALS as u32))
-                / 10u64.pow(token_decimals as u32);
-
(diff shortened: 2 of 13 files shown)


solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -68,13 +73,76 @@ pub mod restaking {
...
+
+            let token_decimals = ctx.accounts.token_mint.decimals;
+
+            let amount_in_sol_decimals = (amount *
+                10u64.pow(SOL_DECIMALS as u32)) /
+                10u64.pow(token_decimals as u32);
+
+            let final_amount_in_sol = ((token_price.price *
+                (amount_in_sol_decimals as i64)) /
+                sol_price.price) as u64;
```
mina86, line 130: You can subtract rather than dividing.  More importantly, I’m not
happy with the i64 and u64 conversions.  Those need to be checked
rather than `as`.  We have overflow checks for other operations but
`as` just interprets the bits.

```suggestion
            let decimals = SOL_DECIMALS - ctx.accounts.token_mint.decimals;
            let amount_in_sol_decimals = amount * 10u64.pow(decimals as u32);
            let amount_in_sol_decimals =
                i64::try_from(amount_in_sol_decimals).unwrap();
            let final_amount_in_sol = token_price.price *
                amount_in_sol_decimals /
                sol_price.price;
            let final_amount_in_sol =
                u64::try_from(final_amount_in_sol).unwrap();
```

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -68,13 +73,76 @@ pub mod restaking {
...
             .whitelisted_tokens
             .iter()
-            .find(|&&token_mint| token_mint == ctx.accounts.token_mint.key())
+            .enumerate()
+            .find_map(|(index, &token_mint)| {
+                if token_mint == ctx.accounts.token_mint.key() {
+                    return Some(index);
+                }
+                None
+            })
```
mina86, line 85: Sounds like you’re looking for [`position`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position).

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -68,13 +73,76 @@ pub mod restaking {
...
             .ok_or_else(|| error!(ErrorCodes::TokenNotWhitelisted))?;
 
-        staking_params.total_deposited_amount += amount as u128;
+        let token_price_update = &ctx.accounts.token_price_update;
+        let sol_price_update = &ctx.accounts.sol_price_update;
+
+        let sol_price_feed_id = SOL_PRICE_FEED_ID.to_string();
```
mina86, line 91: Why convert it to String?

solana/restaking/programs/restaking/src/lib.rs
```diff
@@ -68,13 +73,76 @@ pub mod restaking {
...
+        let token_feed_id = staking_params
+            .token_oracle_addresses
+            .get(token_index)
+            .unwrap_or(&sol_price_feed_id);
+
+        let final_amount_in_sol = if token_feed_id == &sol_price_feed_id {
```
mina86, line 97: Sholudn’t this be `!=`?

dhruvja: oh yes sorry, my bad.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: because the feed id from hex expects &string and if i convert it there then the compiler complains saying use of moved value.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: oh yes right.

solana/restaking/programs/restaking/src/lib.rs
dhruvja: the thing is that the token decimals can be more than solana decimals so i guess we could do smtg like this then
```
let decimals = (SOL_DECIMALS as i64) - (ctx.accounts.token_mint.decimals as i64);
let amount_in_sol_decimals = if decimals < 0 {
     amount / 10u64.pow(abs(decimals))
} else {
     amount * 10u64.pow(decimals)
}
```

Anchor.toml
```diff
@@ -33,9 +33,15 @@ upgradeable = true
...
 rpc_port = 8899
 
 [[test.validator.clone]]
 address = "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s"
+
+[[test.validator.clone]]
+address = "Dpw1EAVrSB1ibxiDQyTAW6Zip3J4Btk2x4SgApQCeFbX"
+
+[[test.validator.clone]]
+address = "7UVimffxr9ow1uXYxsr4LHAcV58mLzhmwaeKvJ1pjLiE"
```
mina86, line 47:
```suggestion
address = "7UVimffxr9ow1uXYxsr4LHAcV58mLzhmwaeKvJ1pjLiE"
```

solana/restaking/programs/restaking/src/lib.rs
mina86: I think what you want is:

```rust
        let token_feed_id = staking_params
            .token_oracle_addresses
            .get(token_index)
            .as_deref()
            .unwrap_or(SOL_PRICE_FEED_ID);
```

dhruvja: oh ya i think `deref` is what was needed.

solana/restaking/programs/restaking/src/lib.rs
mina86: That makes sense, though prefer `i64::from(...)` to `... as i64`.  I also wonder if we should be concerned with rounding? I guess in this instance truncating is the correct approach.

dhruvja:
Well no. We store the calculated value during deposit and use the same value to unstake. The thing is the user will get the same number of tokens, the conversion is only for validator stake so i guess its prolly fine.

mina86:
I’m just wondering about a case if someone deposits 1 FOO at a moment when it’s worth 1000 SOL. So we’re going to record it as 1000 SOL. But say month later it drops to 1 FOO = 1 SOL, but we will still treat this as worth 1000 SOL when calculating total stake and quorum.

dhruvja:
We would just enable stables for now else it's gonna get a bit tricky to update price everytime.

mina86:
OK, that makes more sense. Though even then between 2021 and now price of SOL fluctuated from 10 USD to 250 USD. ;)

dhruvja:
ill keep this in draft since we wont be using it for v1.

PR #335, opened 2024-06-12
solana-ibc: add update_stake call altering stake of multiple validators

dhruvja:
Introduce update_stake method which allows changing stake of multiple
validators in a single go.  It can be called with CPI from the
restaking program only.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/chain.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/chain.rs
@@ -8,6 +8,7 @@ pub use sigverify::ed25519::{PubKey, Signature};
 pub use sigverify::Verifier;
 
 use crate::error::Error;
+use crate::storage::StakeOperation;
 use crate::{events, ibc, storage};
 
 type Result<T = (), E = anchor_lang::error::Error> = core::result::Result<T, E>;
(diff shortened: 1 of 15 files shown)


solana/solana-ibc/programs/solana-ibc/src/chain.rs
```diff
@@ -173,6 +174,42 @@ impl ChainData {
...
     }
 
+    /// Updates multiple validators stake based on the operation
+    ///
+    /// Fails if candidate is not found during remove operation
+    /// or if the subtraction overflows
+    pub fn set_stake_multiple(
+        &mut self,
+        validator_change_in_stake: Vec<(Pubkey, u128)>,
+        operation: StakeOperation,
```
mina86, line 184: I think we can get away with having i128 in the map and then we don’t need StakeOperation.

mina86 (commented):
Also, this PR can be more efficient with https://github.com/ComposableFi/emulated-light-client/pull/337 I can update it once #337 gets merged.

mina86:
I’ve pushed the changes. This is blocked on #337 but once that’s merged this one can go I think. It is somewhat simpler.

dhruvja:
I have merged #337

mina86 (approved):

PR #336, opened 2024-06-12
solana-ibc: make updating chain config by fee collector

dhruvja:
Updating chain config rn is not possible so the config can only be set once during initialization. So adding configurable option which can be executed by the fee collector.

### DIFF
--- a/common/guestchain/src/candidates.rs
+++ b/common/guestchain/src/candidates.rs
@@ -92,6 +92,7 @@ impl<PK: crate::PubKey> Candidates<PK> {
         this
     }
 
+    /// Returns stake held by first `max_validators` candidates.
     pub fn current_head_stake(&self) -> u128 { self.head_stake }
 
     /// Sums stake of the first `count` candidates.
@@ -104,18 +105,14 @@ impl<PK: crate::PubKey> Candidates<PK> {
 
     /// Updates max validators count.
     pub fn update_max_validators(&mut self, max_validators: NonZeroU16) {
-        let total = NonZeroU16::new(self.candidates.len() as u16).unwrap();
-        let old_count = total.min(self.max_validators);
-        let new_count = total.min(max_validators);
+        let total = self.candidates.len();
+        let old_count = total.min(usize::from(self.max_validators.get()));
+        let new_count = total.min(usize::from(max_validators.get()));
         self.max_validators = max_validators;
         if old_count != new_count {
             self.changed = true;
-            self.head_stake = self
-                .candidates
-                .iter()
-                .take(u16::from(max_validators) as usize)
-                .map(|c| c.stake.get())
-                .sum();
+            self.head_stake =
+                Self::sum_head_stake(max_validators, &self.candidates);
         }
     }
 
--- a/common/guestchain/src/candidates.rs
+++ b/common/guestchain/src/candidates.rs
@@ -100,6 +100,11 @@ impl<PK: crate::PubKey> Candidates<PK> {
             .fold(0, |sum, c| sum.checked_add(c.stake.get()).unwrap())
     }
 
+    /// Updates max validators count.
+    pub fn update_max_validators(&mut self, max_validators: NonZeroU16) {
+        self.max_validators = max_validators;
+    }
+
     /// Returns top validators if changed since last time changed flag was
     /// cleared.
     ///
--- a/common/guestchain/src/candidates.rs
+++ b/common/guestchain/src/candidates.rs
@@ -92,9 +92,7 @@ impl<PK: crate::PubKey> Candidates<PK> {
         this
     }
 
-    pub fn current_head_stake(&self) -> u128 {
-        self.head_stake
-    }
+    pub fn current_head_stake(&self) -> u128 { self.head_stake }
 
     /// Sums stake of the first `count` candidates.
     fn sum_head_stake(count: NonZeroU16, candidates: &[Candidate<PK>]) -> u128 {
@@ -106,7 +104,19 @@ impl<PK: crate::PubKey> Candidates<PK> {
 
     /// Updates max validators count.
     pub fn update_max_validators(&mut self, max_validators: NonZeroU16) {
+        let total = NonZeroU16::new(self.candidates.len() as u16).unwrap();
+        let old_count = total.min(self.max_validators);
+        let new_count = total.min(max_validators);
         self.max_validators = max_validators;
+        if old_count != new_count {
+            self.changed = true;
+            self.head_stake = self
+                .candidates
+                .iter()
+                .take(u16::from(max_validators) as usize)
+                .map(|c| c.stake.get())
+                .sum();
+        }
     }
 
     /// Returns top validators if changed since last time changed flag was
--- a/common/guestchain/src/candidates.rs
+++ b/common/guestchain/src/candidates.rs
@@ -128,60 +128,68 @@ impl<PK: crate::PubKey> Candidates<PK> {
 
     /// Adds a new candidates or updates existing candidate’s stake.
     ///
-    /// If `stake` is zero, removes the candidate from the set.
-    pub fn update(
+    /// The `new_stake_fn` callback takes existing candidate or `None` (if
+    /// candidate with given `pubkey` doesn’t exist) as the argument and returns
+    /// the new stake for that candidate (or for a new candidate).  If the new
+    /// stake is zero, the candidate is removed.
+    pub fn update<F, E>(
         &mut self,
         cfg: &crate::Config,
         pubkey: PK,
-        stake: u128,
-    ) -> Result<(), UpdateCandidateError> {
-        match NonZeroU128::new(stake) {
-            None => self.do_remove(cfg, &pubkey),
-            Some(stake) if stake < cfg.min_validator_stake => {
+        new_stake_fn: F,
+    ) -> Result<(), E>
+    where
+        F: FnOnce(Option<&Candidate<PK>>) -> Result<u128, E>,
+        E: From<UpdateCandidateError>,
+    {
+        let pos = self.candidates.iter().position(|el| el.pubkey == pubkey);
+        let stake = new_stake_fn(pos.map(|pos| &self.candidates[pos]))?;
+        let res = if let Some(stake) = NonZeroU128::new(stake) {
+            if stake < cfg.min_validator_stake {
                 Err(UpdateCandidateError::NotEnoughValidatorStake)
+            } else {
+                self.do_update(cfg, pos, Candidate { pubkey, stake })
             }
-            Some(stake) => self.do_update(cfg, Candidate { pubkey, stake }),
-        }
+        } else if let Some(pos) = pos {
+            self.do_remove(cfg, pos)
+        } else {
+            Ok(())
+        };
+        self.debug_verify_state();
+        res.map_err(E::from)
     }
 
     /// Adds a new candidates or updates existing candidate’s stake.
     fn do_update(
         &mut self,
         cfg: &crate::Config,
+        old_pos: Option<usize>,
         candidate: Candidate<PK>,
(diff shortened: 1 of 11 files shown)


common/guestchain/src/manager.rs
```diff
@@ -168,6 +181,43 @@ impl<PK: crate::PubKey> ChainManager<PK> {
         self.pending_block.as_ref()
     }
 
+    pub fn update_config(
```
mina86, line 184: I think I’d actually make it a method of Config, i.e. Config::update.
Then UpdateConfigError could also be defined in config.rs.  And then
this method would just be

```rust
 pub fn update_config(
        &mut self,
        update: UpdateChainConfigPayload,
    ) -> Result<(), UpdateConfigError> {
        self.config.update(update)
    }
```

Not sure if that’s better, but it would move code out of manager.rs
into config.rs resulting in better balance. ;)

dhruvja: ya i think that should be possible.

common/guestchain/src/manager.rs
```diff
@@ -106,6 +107,18 @@ pub enum AddSignatureError {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum UpdateConfigError {
+    /// Minimum validators are more than existing
+    MinValidatorsHigherThanExisting,
+    /// Maximum validators are less than existing
+    MaxValidatorsLowerThanExisting,
+    /// Total Stake is less than existing
+    MinTotalStakeLowerThanExisting,
+    /// Quorum Stake is less than existing
+    MinQuorumStakeLowerThanExisting,
```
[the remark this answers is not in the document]
dhruvja: for example, if the existing validators are 6 and we increase minimum validators to 7, then it doesnt make sense right?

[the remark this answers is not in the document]
mina86: OK, but then the code doesn’t change the current number of validators but rather current configured minimum.  So if minimum is currently 5 but we have 10 validators it’s fine to increase the number to 7.

As to whether it makes sense, in general it is possible that chain is in such invalid state. This can happen when genesis block doesn’t match the configuration. The chain will work fine even if there’s fewer validators than minmum (it’s just that no one will be able to leave). I guess we can check for the condition but then we need to look at actual number of validators and also same with stake.

common/guestchain/src/manager.rs
```diff
@@ -168,6 +181,43 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
+            if min_validators > self.config.min_validators {
+                return Err(UpdateConfigError::MinValidatorsHigherThanExisting);
+            }
+            self.config.min_validators = min_validators;
+        }
+        if let Some(max_validators) = config_payload.max_validators {
+            if max_validators < self.config.max_validators {
+                return Err(UpdateConfigError::MaxValidatorsLowerThanExisting);
+            }
+            self.config.max_validators = max_validators;
```
mina86, line 198: By the way, there’s also self.candidates.max_validators that needs to be updated.

common/guestchain/src/config.rs
```diff
@@ -91,3 +91,13 @@ pub struct Config {
...
 }
+
+#[derive(Clone, Debug, borsh::BorshSerialize, borsh::BorshDeserialize)]
+pub struct UpdateChainConfigPayload {
+    pub min_validators: Option<NonZeroU16>,
+    pub max_validators: Option<NonZeroU16>,
+    pub min_validator_stake: Option<NonZeroU128>,
+    pub min_total_stake: Option<NonZeroU128>,
+    pub min_quorum_stake: Option<NonZeroU128>,
+    pub min_block_length: Option<crate::height::HostDelta>,
```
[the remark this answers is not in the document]
dhruvja: oh yes right, just missed it.

common/guestchain/src/manager.rs
dhruvja: got it

common/guestchain/src/manager.rs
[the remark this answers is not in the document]
dhruvja: oh yes right right.

common/guestchain/src/config.rs
```diff
@@ -91,3 +91,104 @@ pub struct Config {
...
+        }
+        if let Some(min_total_stake) = config_payload.min_total_stake {
+            if u128::from(min_total_stake) > head_stake {
+                return Err(UpdateConfigError::MinTotalStakeHigherThanExisting);
+            }
+            if min_total_stake < self.min_quorum_stake {
+                return Err(
+                    UpdateConfigError::MinTotalStakeHigherThanMinQuorumStake,
+                );
+            }
```
mina86, line 172: Similarly here, this sneeds to be checked outside of the other `if let` checks after min_quorum_stake and min_total_stake have been updated.

common/guestchain/src/manager.rs
```diff
@@ -494,4 +513,51 @@ fn test_generate() {
         false,
     )
     .unwrap();
+
+    let update_chain_config = UpdateConfig {
+        min_validators: Some(
+            NonZeroU16::new((mgr.validators().len() + 1) as u16).unwrap(),
+        ),
```
mina86, line 533:
```suggestion
        min_validators: NonZeroU16::new((mgr.validators().len() + 1) as u16),
```
is also fine. Up to you if you want to be more explicit or more concise.

common/guestchain/src/config.rs
```diff
@@ -91,3 +91,104 @@ pub struct Config {
...
+                return Err(UpdateConfigError::MinValidatorsHigherThanExisting);
+            }
+            self.min_validators = min_validators;
+        }
+        if let Some(max_validators) = config_payload.max_validators {
+            if max_validators < self.min_validators {
+                return Err(
+                    UpdateConfigError::MaxValidatorsCannotBeLowerThanMin,
+                );
+            }
```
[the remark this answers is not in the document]
dhruvja: well we just need to make sure that new minimum is less than the new maximum right? so we could prolly have smtg like this inside `min_validators` condn
```rs
if let Some(min_validators) = config_payload.min_validators {
            if min_validators > NonZeroU16::new(total_validators).unwrap() {
                return Err(UpdateConfigError::MinValidatorsHigherThanExisting);
            }
            if let Some(max_validators) = config_payload.max_validators {
                if max_validators < min_validators {
                    return Err(
                        UpdateConfigError::MaxValidatorsCannotBeLowerThanMin,
                    );
                }
            };
            self.min_validators = min_validators;
        }
```

common/guestchain/src/config.rs
dhruvja: got it.

common/guestchain/src/config.rs
```diff
@@ -91,3 +91,104 @@ pub struct Config {
...
+
+impl Config {
+    pub fn update(
+        &mut self,
+        head_stake: u128,
+        total_validators: u16,
+        config_payload: UpdateConfig,
+    ) -> Result<(), UpdateConfigError> {
+        if let Some(min_validators) = config_payload.min_validators {
+            if min_validators > NonZeroU16::new(total_validators).unwrap() {
```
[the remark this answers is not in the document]
dhruvja: gotcha

common/guestchain/src/candidates.rs
```diff
@@ -27,7 +27,7 @@ pub struct Candidates<PK> {
     changed: bool,
 
     /// Sum of the top `max_validators` stakes.
-    head_stake: u128,
+    pub head_stake: u128,
```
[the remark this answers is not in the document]
dhruvja: right

common/guestchain/src/config.rs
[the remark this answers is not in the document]
mina86: Sure, but it’s easier to just unconditionally check afterwards.

common/guestchain/src/config.rs
```diff
@@ -91,3 +91,118 @@ pub struct Config {
...
+    ///
+    /// If minimum quorum stake is higher than existing total stake, then
+    /// blocks would never get finalized until more stake is added and quorum
+    /// stake is less than head stake.
+    MinQuorumStakeHigherThanTotalStake,
+}
+
+impl Config {
+    pub fn update(
+        &mut self,
```
mina86, line 142: Actually, it just occurred to me that we need to leave `self`
unchanged on an error.  So inside of the function I think we need to
operate on the copy and commit only once everything is done:

```
    pub fn update(&mut self, ...) -> Result<(), UpdateConfigError> {
        let mut this = self.clone();
        ...
        *self = this
    }
```

Alternatively have an `pub fn updated(self, ...) -> Result<Self, ...>`
method and caller would do `self.config =
self.config.clone().update(...)?;`.

common/guestchain/src/candidates.rs
```diff
@@ -100,6 +100,11 @@ impl<PK: crate::PubKey> Candidates<PK> {
             .fold(0, |sum, c| sum.checked_add(c.stake.get()).unwrap())
     }
 
+    /// Updates max validators count.
+    pub fn update_max_validators(&mut self, max_validators: NonZeroU16) {
+        self.max_validators = max_validators;
```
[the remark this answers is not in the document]
dhruvja: makes sense.

common/guestchain/src/config.rs
dhruvja: isnt it atomic?

dhruvja: like if it fails anywhere in between, the changes would be reverted.

common/guestchain/src/config.rs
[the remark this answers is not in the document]
dhruvja: well i feel the above is much simpler, innit?

common/guestchain/src/config.rs
mina86: If you’re talking Solana transaction sure, but this is a type which is blockchain-agnostic and could be used outside of blockchain. Because of that I’d rather it didn’t assume that operation get reverted.

common/guestchain/src/config.rs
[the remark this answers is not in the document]
mina86: Well, it has multiple nested `if`s and `if let Some(max_validators) = ...` is repeated twice so I’d argue is more complex.

common/guestchain/src/manager.rs
```diff
@@ -507,4 +526,49 @@ fn test_generate() {
...
+    assert_eq!(
+        Err(UpdateConfigError::MinValidatorsHigherThanExisting),
+        mgr.update_config(update_chain_config)
+    );
+
+    let update_chain_config = UpdateConfig {
+        min_validators: None,
+        max_validators: NonZeroU16::new(u16::from(config.max_validators) - 1),
+        min_validator_stake: None,
+        min_total_stake: Some(NonZeroU128::new(total_stake + 2).unwrap()),
```
mina86, line 549: Same here and below you could get rid of Some and unwrap.

common/guestchain/src/config.rs
dhruvja: ya right, lemme change that then

common/guestchain/src/manager.rs
dhruvja: gotcha

common/guestchain/src/config.rs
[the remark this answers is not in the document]
dhruvja: well even if u do outside, u would need a condition to get the value inside the option right?

mina86 (approved):

common/guestchain/src/config.rs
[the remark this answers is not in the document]
mina86: I’ve restructured the code. In the original when we were changing `self` it was enough to just verify values in `self` once they are changed. Now this can be done by unwrapping the old value.

PR #337, opened 2024-06-12
guestchain: change ChainManager::update_candidate to take a callback

mina86:
Change the ChainManager::update_candidate such that rather than the new stake it takes a callback function that returns the new stake. This allows modifications to be implemented without having to look up the candidate twice (once to read existing value an the other time to set the new value).


dhruvja (approved):
LGTM

PR #339, opened 2024-06-13
validator: generate block only if the conditions are passed

dhruvja:
Updating the validator client to only generate block if the conditions are satisfied. The validations are done before the call is being made to reduce the unnecessary calls to the program.

Reducing the Compute units to 150k for signing the block since as of now with 24 validators, a maximum of only 70k CU are consumed. The extra CU are requested in case the validator set increases.

### DIFF
--- a/common/guestchain/src/manager.rs
+++ b/common/guestchain/src/manager.rs
@@ -236,10 +236,10 @@ impl<PK: crate::PubKey> ChainManager<PK> {
             return Err(GenerateError::BlockTooYoung);
         }
 
-        let next_epoch = self.validate_generate_next_epoch(host_height);
+        let next_epoch = self.maybe_generate_next_epoch(host_height);
         let age =
             host_timestamp.get().saturating_sub(self.header.timestamp_ns.get());
-        if next_epoch &&
+        if next_epoch.is_none() &&
             !force &&
             state_root == &self.header.state_root &&
             age < self.config.max_block_age_ns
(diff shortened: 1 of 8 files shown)


common/guestchain/src/manager.rs
```diff
@@ -264,6 +278,18 @@ impl<PK: crate::PubKey> ChainManager<PK> {
         })
     }
 
+    fn validate_generate_next_epoch(
```
mina86: `should_generate_next_epoch`?

common/guestchain/src/manager.rs
```diff
@@ -264,6 +278,18 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
+        &self,
+        host_height: crate::HostHeight,
+    ) -> bool {
+        if !host_height
+            .check_delta_from(self.epoch_height, self.config.min_epoch_length)
+        {
+            return false;
+        }
+        true
```
mina86:
```suggestion
        host_height
            .check_delta_from(self.epoch_height, self.config.min_epoch_length)
```

common/guestchain/src/manager.rs
```diff
@@ -233,6 +219,36 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
+            self.header.host_height,
+            self.config.min_block_length,
+        ) {
+            return Err(GenerateError::BlockTooYoung);
+        }
+
+        let next_epoch = self.validate_generate_next_epoch(host_height);
+        let age =
+            host_timestamp.get().saturating_sub(self.header.timestamp_ns.get());
+        if next_epoch &&
```
mina86: This is inverted, right? Should be `!next_epoch`.

mina86 (commented):
LGTM except for the checking logic which I think is inverted.  Let’s also get #340 merged first so we can remove `force` from here.

common/guestchain/src/manager.rs
dhruvja: i removed the method and replaced it to take an immutable ref instead.

common/guestchain/src/manager.rs
dhruvja: i reverted so i think it should be good as before.

common/guestchain/src/manager.rs
```diff
@@ -233,6 +219,36 @@ impl<PK: crate::PubKey> ChainManager<PK> {
...
+            return Err(GenerateError::HasPendingBlock);
+        }
+        if !host_height.check_delta_from(
+            self.header.host_height,
+            self.config.min_block_length,
+        ) {
+            return Err(GenerateError::BlockTooYoung);
+        }
+
+        let next_epoch = self.maybe_generate_next_epoch(host_height);
```
mina86, line 239: I don’t like this approach because it generates the next epoch twice. Perhaps change validate_generate_next to return the epoch?

mina86 (approved):

common/guestchain/src/manager.rs
dhruvja: yes that could work

mina86 (approved):

PR #345, opened 2024-06-15
Update Anchor and Solana dependencies

mina86:
Anchor 0.30 changed how IDL is generated.  Unfortunately that change lead to build failures since now types need to implement IdlBuild trait (which our custom types and more crucially IBC types do not implement).  Fixing this would be rather involved as we’d have to wrap all foreign types leading to uglier code throughout.

Since we don’t actually use the generated IDL, rather than fixing all that simply build Anchor programs without it.


dhruvja:
We get `Access violation in unknown section at address 0x9a of size 1` when we use 1.18 version of solana program.

mina86:
Anchor 0.30 requires Solana 1.18 (they claim it doesn’t but anchor-spl transitively depends on package which requires solana-program 1.18) so let’s leave this for now.

PR #346, opened 2024-06-15
restaking-v2: added restaking program with tests

dhruvja:
In this restaking model, the deposit is equally distributed to all the validators and same during withdrawal. The validators have an option to be part of this model. The validators would have to bond a minimum amount to be part of this new restaking program.


solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+        let cpi_ctx = CpiContext::new_with_signer(
+            ctx.accounts.token_program.to_account_info(),
+            mint_to_ix,
+            seeds,
+        );
+
+        anchor_spl::token::mint_to(cpi_ctx, amount)?;
+
+        // Call guest chain program to update the stake equally
+        let stake_per_validator = amount / common_state.validators.len() as u64;
```
mina86, line 95: amount may be indivisible by the number of validators. We may want to add a reminder to a random validator.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+            from: ctx.accounts.staker_receipt_token_account.to_account_info(),
+            authority: ctx.accounts.staker.to_account_info(),
+        };
+
+        let cpi_ctx = CpiContext::new(ctx.accounts.token_program.to_account_info(), burn_ix);
+
+        anchor_spl::token::burn(cpi_ctx, amount)?;
+
+        // Call guest chain program to update the stake equally
+        let stake_per_validator = (amount / common_state.validators.len() as u64) as i128;
```
mina86, line 172: Similarly here, we may en up reducing stake by less than the amount we’ve burnt. Perhaps round down the amount?

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+            .iter()
+            .any(|token_mint| staking_params.whitelisted_tokens.contains(token_mint));
+
+        if contains_mint {
+            return Err(error!(ErrorCodes::TokenAlreadyWhitelisted));
+        }
+
+        staking_params
+            .whitelisted_tokens
+            .append(&mut new_token_mints.as_slice().to_vec());
```
mina86, line 263:
```suggestion
            .extend_from_slice(new_token_mints.as_slice());
```

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+            .iter()
+            .any(|validator| staking_params.validators.contains(validator));
+
+        if contains_validator {
+            return Err(error!(ErrorCodes::ValidatorAlreadyAdded));
+        }
+
+        staking_params
+            .validators
+            .append(&mut new_validators.as_slice().to_vec());
```
mina86, line 288: Ditto.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
dhruvja: i think that makes sense.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
dhruvja: yes

solana/restaking-v2/programs/restaking-v2/src/lib.rs
dhruvja: i think we should add the remainder to a particular validator because during withdrawal we might be withdrawing the remainder from a different validator.

dhruvja: not that the remainder value would be high, but i think it can cause underflow during withdrawal. The other option might be that if it causes underflow, we remove it from a different validator but then we need to deserialize the chain data on restaking layer which i dont think is really worth it.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -31,7 +37,10 @@ pub mod restaking_v2 {
         let common_state = &mut ctx.accounts.common_state;
 
         common_state.admin = ctx.accounts.admin.key();
-        common_state.whitelisted_tokens = whitelisted_tokens;
+        common_state.whitelisted_tokens = whitelisted_tokens
+            .iter()
+            .map(|token_mint| token_mint.into())
+            .collect::<Vec<StakeToken>>();
```
mina86, line 43: You probably want `into_iter` here.

```suggestion
        common_state.whitelisted_tokens = whitelisted_tokens
            .into_iter()
            .map(StakeToken::from)
            .collect();
```

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -145,6 +174,14 @@ pub mod restaking_v2 {
     pub fn withdraw(ctx: Context<Withdraw>, amount: u64) -> Result<()> {
         let common_state = &mut ctx.accounts.common_state;
 
+        let stake_token_mint = &ctx.accounts.token_mint.key();
+
+        let whitelisted_token = common_state
+            .whitelisted_tokens
+            .iter()
+            .find(|&x| &x.address == stake_token_mint)
```
mina86, line 182:
```suggestion
            .find(|x| &x.address == stake_token_mint)
```

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -312,6 +372,123 @@ pub mod restaking_v2 {
...
+        let sol_price_feed = &ctx.accounts.sol_price_feed;
+
+        let token_mint = ctx.accounts.token_mint.key();
+
+        let validators = common_state.validators.clone();
+
+        let staked_token = common_state
+            .whitelisted_tokens
+            .iter_mut()
+            .find(|whitelisted_token| whitelisted_token.address == token_mint);
```
mina86, line 389:
```suggestion
            .find(|whitelisted_token| whitelisted_token.address == token_mint)
            .ok_or_else(|| error!(ErrorCodes::InvalidTokenMint))?;
```

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -312,6 +372,123 @@ pub mod restaking_v2 {
...
+
+        if let Some(staked_token) = staked_token {
+            if let Some(token_feed_id) = staked_token.oracle_address.as_ref() {
```
mina86, line 392:
```suggestion
        let token_feed_id = staked_token
            .oracle_address
            .as_ref()
            .ok_or_else(|| error!(ErrorCodes::OracleAddressNotFound))?;
```

solana/restaking-v2/programs/restaking-v2/src/lib.rs
```diff
@@ -312,6 +372,123 @@ pub mod restaking_v2 {
...
+                            maximum_age_in_sec,
+                            &feed_id,
+                        )?;
+                    (token_price, sol_price)
+                };
+
+                let token_decimals = ctx.accounts.token_mint.decimals;
+
+                let amount_in_sol_decimals = 10u64.pow(SOL_DECIMALS as u32) /
+                    10u64.pow(token_decimals as u32);
```
mina86, line 423: This may be zero if token_decimals is greater than sol_decimals.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
dhruvja: as per ur analysis, using lazy evaluation is better.

solana/restaking-v2/programs/restaking-v2/src/lib.rs
dhruvja: storing the value by multiplying with exponent so that we dont miss the fractions.

PR #348, opened 2024-06-18
solana-ibc: Unwrap sol when unescrowing wrapped sol.

dhruvja:
Whenever someone sends SOL to Solana, they get wrapped sol instead of native SOL which is unusable if the wallet didnt have SOL previously. So whenever someone send SOL from solana, the escrow account which holds wsol gets closed and all the lamports are sent to the `mint_authority` account which is the owner of the escrow account. When we unescrow the wrapped sol, we debit it from the `mint_authority` account instead of transferring from the `escrow` account. This will allow users to receive native SOL on solana instead of wrapped sol. Though when users deposit, they need to have wrapped SOL which can be a prior instruction to convert SOL to wSOL.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/tests.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/tests.rs
@@ -15,12 +15,12 @@ use anchor_client::solana_sdk::signature::{
 };
 use anchor_client::solana_sdk::transaction::Transaction;
 use anchor_client::{Client, Cluster};
-use anchor_lang::solana_program::system_instruction::create_account;
 use anchor_lang::AnchorSerialize;
 use anchor_spl::associated_token::get_associated_token_address;
 use anyhow::Result;
 use ibc::apps::transfer::types::msgs::transfer::MsgTransfer;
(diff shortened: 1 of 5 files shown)


solana/solana-ibc/programs/solana-ibc/src/tests.rs
```diff
@@ -575,7 +557,7 @@ fn anchor_test_deliver() -> Result<()> {
     assert_eq!(
         ((account_balance_before.ui_amount.unwrap() -
             account_balance_after.ui_amount.unwrap()) *
-            10_f64.powf(mint_info.decimals.into()))
+            10_f64.powf(9.0))
```
mina86, line 560:
```suggestion
            10_f64.powi(9))
```
would also work; or just hard code the number.

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/tests.rs
dhruvja: makes sense.

PR #349, opened 2024-06-26
solana-ibc: method to update connection delay of specified connection id

dhruvja:
Adding a method which would enable us to change the connection delay period of a specified connection end. This method can only be called by `fee_collector` and would fail if a `connection_id` that doesnt exist is passed.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -129,6 +129,9 @@ pub enum Error {
     /// blocks would never get finalized until more stake is added and quorum
     /// stake is less than head stake.
     MinQuorumStakeHigherThanTotalStake,
+
+    /// When a connection id which is not yet created is updated
+    InvalidConnectionId,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/error.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/error.rs
@@ -129,9 +129,6 @@ pub enum Error {
     /// blocks would never get finalized until more stake is added and quorum
     /// stake is less than head stake.
     MinQuorumStakeHigherThanTotalStake,
-
-    /// When a connection id which is not yet created is updated
-    InvalidConnectionId,
 }
 
 impl Error {
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -536,35 +536,7 @@ pub mod solana_ibc {
             return Err(error!(error::Error::InvalidConnectionId));
         }
 
-        let provable = crate::storage::get_provable_from(
-            &ctx.accounts.trie,
-            &ctx.accounts.fee_collector,
-        )?;
-        let chain = &mut ctx.accounts.chain;
-
-        // Before anything else, try generating a new guest block.  However, if
-        // that fails it’s not an error condition.  We do this at the beginning
-        // of any request.
-        chain.maybe_generate_block(&provable)?;
-
-        let accounts = crate::TransferAccounts {
-            sender: None,
-            receiver: None,
-            token_account: None,
-            token_mint: None,
-            escrow_account: None,
-            mint_authority: None,
-            token_program: None,
-            fee_collector: None,
-        };
-
-        let mut store =
-            crate::storage::IbcStorage::new(crate::storage::IbcStorageInner {
-                private: storage,
-                provable,
-                chain,
-                accounts,
-            });
+        let mut store = crate::storage::from_ctx!(ctx);
 
         let connection_id = ConnectionId::new(connection_id);
         let connection_end = store.connection_end(&connection_id).unwrap();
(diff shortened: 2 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/error.rs
```diff
@@ -129,6 +129,9 @@ pub enum Error {
     /// blocks would never get finalized until more stake is added and quorum
     /// stake is less than head stake.
     MinQuorumStakeHigherThanTotalStake,
+
+    /// When a connection id which is not yet created is updated
+    InvalidConnectionId,
```
mina86, line 134: Would it make sense to use https://docs.rs/ibc/latest/ibc/core/connection/types/error/enum.ConnectionError.html#variant.ConnectionNotFound instead?

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -513,6 +519,75 @@ pub mod solana_ibc {
...
+            receiver: None,
+            token_account: None,
+            token_mint: None,
+            escrow_account: None,
+            mint_authority: None,
+            token_program: None,
+            fee_collector: None,
+        };
+
+        let mut store =
```
mina86, line 561: Uh? Why not storage::from_ctx!?

solana/solana-ibc/programs/solana-ibc/src/error.rs
dhruvja: i think it would.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: well then i had to rename the signer as `sender`, though it really doesnt matter what i name.

mina86: I think it’s better to rename it so we don’t have to repeat the same code.  If we *really* care about the name there might be way to make it an option in the macro, but we probably don’t care enough.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -96,6 +96,12 @@ solana_program::custom_panic_default!();
 
 #[anchor_lang::program]
 pub mod solana_ibc {
+    use std::time::Duration;
+
+    use ::ibc::core::channel::context::SendPacketValidationContext;
+    use ::ibc::core::connection::types::ConnectionEnd;
+    use ::ibc::core::host::types::identifiers::ConnectionId;
+    use ::ibc::core::host::ExecutionContext;
```
mina86, line 104: Those are all re-exported in `ibc` module. I’d just use `ibc::Foo` directly where those types are used.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -513,6 +519,54 @@ pub mod solana_ibc {
...
+        delay_period_in_ns: u64,
+    ) -> Result<()> {
+        let storage = &mut ctx.accounts.storage;
+
+        let connection_id = ConnectionId::new(connection_id_idx);
+
+        // Panic if connection_id doenst exist
+        if storage.connections.len() >= connection_id_idx as usize {
+            return Err(error!(error::Error::ContextError(
+                crate::ibc::ContextError::ConnectionError(
```
mina86, line 539: `crate::` prefix shouldn’t be needed.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -513,6 +519,54 @@ pub mod solana_ibc {
...
+        ctx: Context<UpdateConnectionDelay>,
+        connection_id_idx: u64,
+        delay_period_in_ns: u64,
+    ) -> Result<()> {
+        let storage = &mut ctx.accounts.storage;
+
+        let connection_id = ConnectionId::new(connection_id_idx);
+
+        // Panic if connection_id doenst exist
+        if storage.connections.len() >= connection_id_idx as usize {
```
mina86, line 537: I’m not a fan of `as` casting, though I suppose there’s no obvious way
to replace it.  We could change `connection_id_idx` to `u16` and then
use `usize::from(connection_id_idx)` or do
`usize::try_from(connection_id_idx).unwrap()`.  Not sure if either of
those is actually better.

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -513,6 +519,54 @@ pub mod solana_ibc {
...
+                    crate::ibc::ConnectionError::ConnectionNotFound {
+                        connection_id: connection_id.clone()
+                    }
+                )
+            )));
+        }
+
+        let mut store = crate::storage::from_ctx!(ctx);
+
+        let connection_end = store.connection_end(&connection_id).unwrap();
```
mina86, line 549:
```suggestion
        let connection_end = store.connection_end(&connection_id)?;
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -513,6 +519,54 @@ pub mod solana_ibc {
...
+            connection_end.client_id().clone(),
+            connection_end.counterparty().clone(),
+            connection_end.versions().to_vec(),
+            Duration::from_nanos(delay_period_in_ns),
+        )
+        .unwrap();
+
+        store
+            .store_connection(
+                &::ibc::core::host::types::path::ConnectionPath(connection_id),
```
mina86, line 562:
```suggestion
                &ibc::path::ConnectionPath(connection_id),
```
should work

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/lib.rs
dhruvja: if we use u16, we would need to convert it again to get `connection_id`

mina86: Yes, but you can do it with just `_.into()` or `u64::from(_)`.

dhruvja: but what is the benefit of getting rid of `as` casting. does it affect the performance?

dhruvja: oh i see. so it makes sense to send it as u16 and use `from` and `into` for conversions then.

PR #360, opened 2024-08-01
Pin Rust version in CI to avoid time build failures

mina86:
Unfortunately, Rust 1.80 breaks `time` dependency.  Specifically,
a build fails with the following error:

    error[E0282]: type annotations needed for `Box<_>`
      --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.30/src/format_description/parse/mod.rs:83:9
       |
    83 |     let items = format_items
       |         ^^^^^
    ...
    86 |     Ok(items.into())
       |              ---- type must be known at this point

This shouldn’t be a problem since we can update the dependency.
Unfortunately, when we install Anchor CLI, that still uses old `time`
crate and build of that fails.

For the time being, pin the compiler version in CI to 1.80 to avoid
this build failure.

This commit also updates `time` dependency in case our code is built
with newest Rust.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -9,6 +9,10 @@ on:
       - master
 
 env:
+  # Rust 1.80 breaks `time` dependency.  We have updated *our* dependency but
+  # Anchor CLI 0.29 uses old `time` which doesn’t compile with new Rust.
+  # Because of that, we limit the stable Rust version to 1.79.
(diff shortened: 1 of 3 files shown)


dhruvja (approved):
As long as we are using solana v1.17 to compile, i dont think we would have any issues.

PR #361, opened 2024-08-06
fast-bridge: Add fast bridge contract 

dhruvja:
Has the escrow contract for fast bridge which would be used to escrow the funds of the solver and unescrow when it receives the acknowledgement through IBC.

### DIFF
--- a/solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
+++ b/solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
@@ -1,6 +1,9 @@
-use anchor_spl::token;
+use std::str::FromStr;
+
 use anchor_lang::prelude::*;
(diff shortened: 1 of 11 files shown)


solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
```diff
@@ -0,0 +1,840 @@
...
+
+    pub token_program: Program<'info, Token>,
+
+    // New Intent account addition
+    #[account(
+        mut, 
+        seeds = [b"intent", intent_id.as_bytes()], 
+        bump,
+        constraint = intent.user_in == authority.key().to_string()
+    )]
```
dhruvja, line 607: The `user_in` is the user who escrowed the funds but the caller is the solana-ibc contract or auctioneer.
```suggestion
    #[account(
        mut, 
    )]
```

PR #370, opened 2024-08-20
cf-solana: add bank hash handling plus some refactoring

mina86:
Firstly, add DeltaHashProof type which contains all the information that are needed to calculate delta hash.  One piece of information in it is the accounts delta hash.  As such, this struct together with a trusted bank hash acts as a proof for the accounts delta hash.

Secondly, create AccountHashData wrapper for the account_hash_data field of the AccountProof struct.  This improves type safety and better encapsulates the account information used in hash calculation.

Lastly, fix a few instances where solana-program2 Cargo feature name was used instead of proper solana-program-2.

### DIFF
--- a/common/cf-solana/src/proof.rs
+++ b/common/cf-solana/src/proof.rs
@@ -184,15 +184,83 @@ impl<'a> core::iter::ExactSizeIterator for ProofLevels<'a> {
 impl<'a> core::iter::FusedIterator for ProofLevels<'a> {}
 
 
+/// Data that goes into account’s hash.
+#[derive(Clone, Debug, derive_more::Into, derive_more::AsRef)]
+pub struct AccountHashData(Vec<u8>);
+
+impl AccountHashData {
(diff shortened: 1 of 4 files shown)


dhruvja:
we have to add `solana-program-2` feature in `cf-solana` crate right? i can only see `solana-program` feature.
https://github.com/ComposableFi/emulated-light-client/blob/mpn/d/common/cf-solana/Cargo.toml#L30-L33

mina86:
An optional dependency is implicitly a feature: https://github.com/ComposableFi/emulated-light-client/blob/mpn/d/common/cf-solana/Cargo.toml#L18

dhruvja:
> An optional dependency is implicitly a feature: https://github.com/ComposableFi/emulated-light-client/blob/mpn/d/common/cf-solana/Cargo.toml#L18

oh i see. sounds good then.

dhruvja (approved):

PR #374, opened 2024-08-22
fast-bridge: Emit events for each method.

dhruvja:
Emitting events for all instructions so that they can be parsed from the logs and be acted upon.


dhruvja:
Will `fmt` after review since it results in a huge amount of changes.

solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
```diff
@@ -135,10 +174,22 @@ pub mod bridge_escrow {
...
+        .map_err(|err| {
+            msg!("{}", err);
+            ErrorCode::InvalidEventFormat
+        })?;
+
         Ok(())
     }
 
     // this function is called by Solver
+    #[allow(unused_variables)]
```
mina86, line 192: Why not prefix with underscore?

solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
```diff
@@ -83,16 +99,39 @@ pub mod bridge_escrow {
...
+
+        let current_timestamp = Clock::get()?.unix_timestamp as u64;
+
+        intent.intent_id = intent_id.clone();
         intent.user = user_in;
         intent.token_in = token_in;
         intent.amount_in = amount_in;
-        intent.token_out = token_out;
+        intent.token_out = token_out.clone();
         intent.timeout_timestamp_in_sec = timeout_in_sec;
```
mina86, line 110: Not exactly part of this PR, but is the timeout relative duration or is it a timestamp? This is also what I’ve commented on the other PR.

mina86 (approved):

solana/bridge-escrow/programs/bridge-escrow/src/lib.rs
dhruvja: resolved in the other PR.

PR #375, opened 2024-08-22
fast-bridge: Send required accounts for cross domain transfer

dhruvja:
For cross domain transfers, an acknowledgement is sent from destination to source in form of a token transfer. The packet contains the memo that has the information about unlocking the funds. The source would then parse the memo and unlock the funds to the solver mentioned in the memo. Since we only have token transfers enabled and no cross chain messages, a token needs to be transferred. For this, we create a dummy token owned by the program which mints a token everytime and sends a transfer. The timeout of the transfer is set to infinite so that it always reaches the counterparty chain.

Also the `on_receive_transfer` method accepts a memo which unlocks the funds to the solver. Right now it is being called by the auctioneer but once the hooks are enabled on `solana-ibc` program, this method would then be called by the `solana-ibc` program.

### DIFF
--- a/solana-test.sh
+++ b/solana-test.sh
@@ -1,4 +1,4 @@
-# !/bin/sh
+#!/bin/sh
 set -eux
 solana config set --url http://127.0.0.1:8899
 cd solana/write-account
--- a/solana-test.sh
+++ b/solana-test.sh
@@ -1,15 +1,15 @@
-#!/bin/sh
-# set -eux
-# solana config set --url http://127.0.0.1:8899
-# cd solana/write-account
-# cargo build-sbf
-# cd ../..
-# cd solana/signature-verifier
-# cargo build-sbf
-# cd ../..
-# solana program deploy target/deploy/write.so
-# solana program deploy target/deploy/sigverify.so
-# cargo test  --lib -- --nocapture --include-ignored ::anchor
+# !/bin/sh
+set -eux
+solana config set --url http://127.0.0.1:8899
+cd solana/write-account
+cargo build-sbf
+cd ../..
+cd solana/signature-verifier
+cargo build-sbf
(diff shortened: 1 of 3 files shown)


solana-test.sh
```diff
@@ -1,15 +1,15 @@
...
-# cd solana/write-account
-# cargo build-sbf
-# cd ../..
-# cd solana/signature-verifier
-# cargo build-sbf
-# cd ../..
-# solana program deploy target/deploy/write.so
-# solana program deploy target/deploy/sigverify.so
-# cargo test  --lib -- --nocapture --include-ignored ::anchor
+# !/bin/sh
```
mina86, line 1:
```suggestion
#!/bin/sh
```

solana/bridge-escrow/programs/bridge-escrow/src/tests.rs
```diff
@@ -397,82 +385,128 @@ fn escrow_bridge_program() -> Result<()> {
...
+    .0;
+
+    let new_intent = IntentPayload {
+        intent_id: intent_id.clone(),
+        user_in: user.pubkey(),
+        token_in,
+        amount_in: TRANSFER_AMOUNT,
+        token_out: token_out.to_string(),
+        amount_out: amount_out.to_string(),
+        timeout_timestamp_in_sec: 10000,
```
mina86, line 408: That’s 3 am on 1 of January 1970. If this is meant to represent relative timeout the field should be called `timeout_in_set`. `timestamp` implies this is seconds since Unix epoch.

dhruvja: yes, my bad. It is supposed to be timestamp.

mina86 (approved):

PR #380, opened 2024-08-25
lib: move Rayon wrappers to lib crate

mina86:
The Rayon wrapper methods are not specific to the cf-solana crate.
Move them to lib so that any code in the future can easily use them if
necessary.


common/cf-solana/src/blake3.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Otherwise, the calculation is done by `blake3` crate.
+#[allow(dead_code)]
+pub fn hash(bytes: &[u8]) -> Hash {
+    if cfg!(target_os = "solana-program") &&
+        (cfg!(feature = "solana-program") ||
+            cfg!(feature = "solana-program-2"))
+    {
+        hashv(&[bytes])
+    } else {
+        Hash(::blake3::hash(bytes).into())
```
dhruvja, line 18: the blake3 is blocked on mainnet so i dont think we can use it until they enable it. Ref (https://github.com/anza-xyz/agave/wiki/Feature-Gate-Activation-Schedule#features-are-blocked)

dhruvja (approved):

common/cf-solana/src/blake3.rs
mina86: That is unfortunate. It’s going to be a problem when adding cf-solana to solana-ibc.

PR #381, opened 2024-08-27
fast-bridge: Timeouts

dhruvja:
If the solver is not able to solve the intent, then the escrowed funds can be issued back to the user. In case of single domain intents, the funds just get transferred to the user from the escrow account. But for cross chain intents, a timeout message is sent to the other chain with a flag of `Timeout` and the user who should receive it. 

The intent should store the user on source chain so that during timeout, we know whom to refund the funds to. For the memo, it is the same format and only the receiver is changed to user address instead of solver.


doradelta (approved):
ok

PR #385, opened 2024-08-29
trie-geyser: add RPC server to the plugin

mina86:
Add an HTTP JSON RPC server to the witnessed-trie-geyser-plugin so
that relayer can connect to it to gather all the information necessary
for generating the proofs.

Furthermore, as minor housekeeping, rename solana/trie-geyser to
solana/trie-geyser-plugin so that common/trie-geyser is available for
the code common between the plugin and client.


solana/trie-geyser-plugin/config.json
```diff
@@ -0,0 +1,6 @@
+{
+	"libpath": "target/release/libwitnessed_trie_geyser_plugin.so",
```
dhruvja, line 2:
```suggestion
	"libpath": "target/release/libwitnessed_trie_geyser_plugin.dylib",
```

mina86: Looks like Mac thing? Looks like there’s no one thing that would work for Linux and Mac. I’ll update the README.

dhruvja: oh i see, maybe.

solana/trie-geyser-plugin/src/worker.rs
```diff
@@ -128,27 +154,31 @@ impl Worker {
...
                     return;
                 }
             }
         };
 
-        // If the witness account is not in collection of changed accounts,
-        // don’t do anything.
+        // If the trie or witness accounts are not in collection of changed
+        // accounts, don’t do anything.
         if !entry.accounts.contains_key(&self.config.witness_account) {
```
dhruvja, line 168: Shouldnt we calculate the bank hash when epoch changes?

dhruvja (approved):

solana/trie-geyser-plugin/src/worker.rs
mina86: Since we don’t have guest chain there are no epoch changes happening.  With the witness account we only care about the witness account changing.  The only thing it witnesses is the trie and that is only affected by writes to the trie.

PR #389, opened 2024-09-06
solana-trie: make witness generic type

mina86:
Allow customising TrieAccount through a generic witness type.  The main use case is making it TrieAccount Send and Sync by removing support for the witness.  This is needed by, for example, relayer.


dhruvja (approved):
LGTM

PR #391, opened 2024-09-06
solana-ibc: return status as active if client state not found

dhruvja:
If the client state is not found when finding the status, it means that a new client is getting created in which case, we should return the status as active unless the client state is already frozen ( which is checked before ).

### DIFF
--- a/common/cf-solana/src/client/impls.rs
+++ b/common/cf-solana/src/client/impls.rs
@@ -168,7 +168,6 @@ impl ibc::ClientStateCommon for ClientState {
     }
 }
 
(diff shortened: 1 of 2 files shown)


common/cf-solana/src/client/impls.rs
```diff
@@ -168,7 +168,6 @@ impl ibc::ClientStateCommon for ClientState {
     }
 }
 
-
```
mina86, line 171: Again, bit of diff noise.

mina86 (approved):

PR #392, opened 2024-09-07
light-clients: update revision number for cf-guest and cf-solana to 1

dhruvja:
This is to increase compatibility with other libraries which fail to
properly decode proto3 message with missing revision number field
(missing fields should be decoded as zero in proto3).

### DIFF
--- a/common/cf-guest/src/client/impls.rs
+++ b/common/cf-guest/src/client/impls.rs
@@ -270,7 +270,7 @@ where
         ctx.store_consensus_state(
             ibc::path::ClientConsensusStatePath::new(
                 client_id.clone(),
-                0,
+                1,
                 u64::from(self.latest_height),
             ),
             consensus_state.into(),
--- a/common/cf-solana/src/client/impls.rs
+++ b/common/cf-solana/src/client/impls.rs
@@ -188,7 +188,7 @@ where
         ctx.store_consensus_state(
             ibc::path::ClientConsensusStatePath::new(
                 client_id.clone(),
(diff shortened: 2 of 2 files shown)


mina86:
I don’t follow, why do we need this?

dhruvja:
well we used revision number as 1 everywhere since ibc-go used to omit it if it is 0. In our case it doesnt really matter what we use a revision number but i think, using 1 everywhere seems like a good option to maintain backward compatibility.

dhruvja:
if we make it to 0, we wouldnt be able to establish a connection with cosmos since it omits the field if the value is 0.

mina86:
Yes, that’s how protobuf works. In proto3 zero fields are not encoded but then when decoding zero is implied. So Cosmos should decode it as zero just fine.

dhruvja:
but apparently it doesnt, i am not sure why. When we were having revision height as 0, we were getting missing field issues on ibc-go.

mina86 (approved):
OK, though now the rollout is potentially tricky, no? This needs to be updated in the solana-ibc and on Picasso at the same time, no?

dhruvja:
well no, since picasso uses cf-guest from `composable-ibc`.

mina86:
But that uses this cf-guest as dependency, no?

dhruvja:
the cf-guest light client on picasso doesnt implement `ClientStateExecution` trait so we wont have an issue over there.

mina86:
LGTM

PR #393, opened 2024-09-07
solana-ibc: use solana height and timestamp for cf-guest host metadata

dhruvja:
since the cf-guest light client would exist on rollup, the host metadata is the solana slot and the timestamp instead of those fetched from the guest chain since the rollup wouldnt have it.

But if the connection is established to another svm where the guest chain is used, then the above should be reverted.


solana/solana-ibc/programs/solana-ibc/src/client_state.rs
```diff
@@ -236,14 +237,20 @@ impl cf_guest::CommonContext<sigverify::ed25519::PubKey>
...
+                description: e.to_string(),
             })?;
 
-        let height = u64::from(self.borrow().chain.head()?.block_height);
-        let height = ibc::Height::new(1, height)?;
+        let slot = clock.slot;
+        let timestamp_sec = clock.unix_timestamp as u64;
+
+        let timestamp =
+            ibc::Timestamp::from_nanoseconds(timestamp_sec * 10u64.pow(9))
```
mina86, line 249:
```suggestion
            ibc::Timestamp::from_nanoseconds(timestamp_sec * 10u64.ipow(9))
```

mina86:
I think this needs to be conditional. If we’re running on Solana we have chain account and we use time from there. But if witness account is enabled, we don’t have chain account and then we read data directly from blockchain.

solana/solana-ibc/programs/solana-ibc/src/client_state.rs
dhruvja: we would need `num` dependency if we wanna use `ipow` right?

mina86: Sorry, ignore this.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -43,12 +49,41 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
...
+                        description: e.to_string(),
+                    })?;
+
+            let slot = clock.slot;
+            let height = ibc::Height::new(1, slot)?;
+            return Ok(height);
+        }
         let height = u64::from(self.borrow().chain.head()?.block_height);
         let height = ibc::Height::new(1, height)?;
         Ok(height)
```
mina86, line 66: In general, prefer `cfg!` macro to conditional compilation:

```suggestion
        let height = if cfg!(feature = "witness") {
            anchor_lang::solana_program::sysvar::clock::Clock::get()
                .map_err(|e| ibc::ClientError::ClientSpecific {
                    description: e.to_string(),
                })?.slot
        } else {
            u64::from(self.borrow().chain.head()?.block_height)
        };
        ibc::Height::new(1, height)
```

It often makes the code easier to read and makes build failure more
obvious.  With conditional compilation, you need to build with all
combination of features to make sure everything compiles correctly.
With `cfg!` macro, everything is always well formed.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -43,12 +49,41 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
...
+                timestamp_sec * 10u64.pow(9),
+            )
+            .map_err(|e| {
+                ibc::ClientError::ClientSpecific { description: e.to_string() }
+                    .into()
+            });
+        }
         let timestamp = self.borrow().chain.head()?.timestamp_ns.get();
         ibc::Timestamp::from_nanoseconds(timestamp).map_err(|err| {
             ibc::ClientError::Other { description: err.to_string() }.into()
```
mina86, line 89:
```suggestion
        let timestamp_ns = if cfg!(feature = "witness") {
            anchor_lang::solana_program::sysvar::clock::Clock::get()
                .map_err(|e| ibc::ClientError::ClientSpecific {
                    description: e.to_string(),
                })?.clock.unix_timestamp as u64 * 10u64.pow(9)
        } else {
            self.borrow().chain.head()?.timestamp_ns.get()
        };
        ibc::Timestamp::from_nanoseconds(timestamp).map_err(|err| {
            ibc::ClientError::ClientSpecific { description: err.to_string() }.into()
```

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -2,14 +2,15 @@ use std::str::FromStr;
...
+#[cfg(feature = "witness")]
+use anchor_lang::prelude::SolanaSysvar;
 use lib::hash::CryptoHash;
 
 use crate::client_state::AnyClientState;
 use crate::consensus_state::AnyConsensusState;
 use crate::ibc::{self, ConsensusState};
 use crate::storage::{self, IbcStorage};
 
-
```
mina86, line 12: bit of diff noise.

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: i see, that makes sense

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: gotcha

PR #396, opened 2024-10-15
light-clients: make prefix empty for both cf-solana and cf-guest

dhruvja:
Since prefix is expected to be empty, send empty prefixes when verifying proofs.

### DIFF
--- a/common/cf-guest/src/client/impls.rs
+++ b/common/cf-guest/src/client/impls.rs
@@ -199,7 +199,7 @@ impl<PK: PubKey> ibc::ClientStateCommon for ClientState<PK> {
     /// See [`proof::verify`] for documentation of the proof format.
     fn verify_membership(
         &self,
-        prefix: &ibc::CommitmentPrefix,
+        _prefix: &ibc::CommitmentPrefix,
         proof: &ibc::CommitmentProofBytes,
         root: &ibc::CommitmentRoot,
         path: ibc::path::Path,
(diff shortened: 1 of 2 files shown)


mina86:
Yes, but the point is that the prefix sent from relayer is empty and we want to verify that.  If it’s not empty than relayer sends wrong proof.

dhruvja:
The old ibc doesnt support empty prefixes but i guess we can change it.
https://github.com/ComposableFi/composable-ibc/blob/fa46171b7972d38de5da8b29a999420fe2e580cb/ibc/modules/src/core/ics23_commitment/commitment.rs#L157

mina86:
You can create them via `CommitmentPrefix::default()`.  It’s only when you try to create it from a vector that it complains.

dhruvja:
oh ya true, didnt notice that. In that case we can close this PR

dhruvja:
@mina86 apparently the old ibc doesnt support empty prefix at all.
I get the following error
```
 error converting message type into domain type: ics02 client error

    Caused by:
        empty prefix
```

common/cf-guest/src/client/impls.rs
```diff
@@ -199,15 +199,15 @@ impl<PK: PubKey> ibc::ClientStateCommon for ClientState<PK> {
...
         &self,
-        prefix: &ibc::CommitmentPrefix,
+        _prefix: &ibc::CommitmentPrefix,
         proof: &ibc::CommitmentProofBytes,
         root: &ibc::CommitmentRoot,
         path: ibc::path::Path,
         value: Vec<u8>,
     ) -> Result {
         let value = Some(value.as_slice());
         proof::verify_for_block(
```
mina86, line 209:
```suggestion
        // TODO: Once IBC is updated everywhere to version which supports empty
        // prefixes, change `&[]` with `prefix.as_bytes()`.
        proof::verify_for_block(
```

mina86 (approved):
_sighs_ Why is IBC so broken? I guess alternative would be to patch ibc in composable-ibc.

PR #397, opened 2024-10-15
cf-solana: get timestamp in ns for cf-solana

dhruvja:
in the `consensus_has_expired` method, the consensus timestamp is fetched in seconds but added to a trusting period which is in nanoseconds creating a mismatch. So fixing this bug by fetching the consensus timestamp in nanoseconds.

### DIFF
--- a/common/cf-solana/src/client/impls.rs
+++ b/common/cf-solana/src/client/impls.rs
@@ -2,6 +2,8 @@ use alloc::string::ToString;
 use alloc::vec::Vec;
 use core::num::NonZeroU64;
 
+use ibc_core_client_context::consensus_state::ConsensusState as _;
+
 use crate::proto::Any;
 use crate::{
     proof, ClientMessage, ClientState, ConsensusState, Header, Misbehaviour,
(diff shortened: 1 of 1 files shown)


common/cf-solana/src/client/impls.rs
```diff
@@ -168,7 +170,6 @@ impl ibc::ClientStateCommon for ClientState {
     }
 }
 
-
```
mina86, line 171: I think you did `cargo fmt` with non-nightly and that introduced a bit of diff noise. ;)

common/cf-solana/src/client/impls.rs
```diff
@@ -486,8 +486,8 @@ impl ClientState {
         host_timestamp: ibc::Timestamp,
     ) -> bool {
         let expiry_ns = consensus
-            .timestamp_sec
-            .get()
+            .timestamp()
+            .nanoseconds()
```
mina86, line 490: So was this a bug before? Since timestamp_sec implies it was in seconds but we didn’t multiply it to get nanoseconds? If this is a bug fix, could you mention that and details of it in the PR description?

mina86 (approved):

common/cf-solana/src/client/impls.rs
dhruvja: its a bug fix. ill mention in the PR

common/cf-solana/src/client/impls.rs
dhruvja: oh yes maybe

PR #398, opened 2024-10-15
solana-ibc: write local consensus state when witness feature is enabled

dhruvja:
Local consensus state is required for establishing connections. It is required only to establish connection after which it wont be required. So we store the consensus state every time trie changes for a maximum of 64 states.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -402,10 +402,25 @@ pub mod solana_ibc {
         let height = store.borrow().chain.head()?.block_height;
         // height just before the data is added to the trie.
         msg!("Current Block height {}", height);
+        let previous_root = store.borrow().provable.hash().clone();
 
         ::ibc::core::entrypoint::dispatch(&mut store, &mut router, message)
             .map_err(error::Error::ContextError)
             .map_err(move |err| error!((&err)))?;
+
+        #[cfg(feature = "witness")]
+        {
+            let root = store.borrow().provable.hash().clone();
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -402,11 +402,26 @@ pub mod solana_ibc {
...
             .map_err(error::Error::ContextError)
             .map_err(move |err| error!((&err)))?;
 
+        #[cfg(feature = "witness")]
+        {
+            let root = store.borrow().provable.hash().clone();
+            if previous_root != root {
+                msg!("Writing local consensus state");
+                let slot = Clock::get()?.slot;
+                let timestamp = Clock::get()?.unix_timestamp as u64;
```
mina86, line 417:
```suggestion
                let clock = Clock::get()?;
                let slot = clock.slot;
                let timestamp = clock.unix_timestamp as u64;
```

solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -402,11 +402,26 @@ pub mod solana_ibc {
...
+                let storage = &mut store.borrow_mut().private;
+                storage
+                    .add_local_consensus_state(slot, timestamp, root)
+                    .unwrap();
```
mina86, line 421: Could also just chain eliminating the local variable:

```suggestion
                store.borrow_mut()
                    .private;
                    .add_local_consensus_state(slot, timestamp, root)
                    .unwrap();
```

Matter of taste I suppose.

mina86 (approved):

PR #400, opened 2024-10-16
witness: store local consensus state during initialization for witness feature

dhruvja:
When establishing the connection, we need to store the consensus state on chain before the client is stored since the consensus state would be required during `OpenTry` or `OpenAck` based on order of how the connection is established. So storing the consensus state during `initialize` method so that when the client is created, it can use the stored consensus state.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/lib.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/lib.rs
@@ -123,7 +123,7 @@ pub mod solana_ibc {
             let storage = &mut ctx.accounts.storage;
             let clock = Clock::get()?;
             let slot = clock.slot;
-            let timestamp = clock?.unix_timestamp as u64;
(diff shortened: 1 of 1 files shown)


solana/solana-ibc/programs/solana-ibc/src/lib.rs
```diff
@@ -118,6 +118,19 @@ pub mod solana_ibc {
             &ctx.accounts.trie,
             &ctx.accounts.sender,
         )?;
+        #[cfg(feature = "witness")]
+        {
+            let storage = &mut ctx.accounts.storage;
+            let slot = Clock::get()?.slot;
+            let timestamp = Clock::get()?.unix_timestamp as u64;
```
mina86, line 125:
```suggestion
            let clock = Clock::get()?;
            let slot = clock.slot;
            let timestamp = clock?.unix_timestamp as u64;
```

mina86 (approved):

PR #401, opened 2024-10-18
solana-ibc: get host consensus state based on feature

dhruvja:
Host consensus state is required for establishing the connection. So the host consensus state should be returned based on whether the `witness` feature is present or not. If the feature is present, then we return the rollup consensus state else we return the guest consensus state. But the issue is if the `witness` feature is not enabled, it could either be guest or wasm ( which is a wrapper version of guest ). Since the connection with cosmos is already established, we would stick to returning the `guest` consensus state for now.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
@@ -91,9 +91,7 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
                     .find(|cs| cs.0 == height.revision_height())
                     .map(|(_slot, timestamp, trie_root)| {
                         let state = cf_solana::ConsensusState {
-                            trie_root: ibc::CommitmentRoot::from_bytes(
-                                trie_root.as_slice(),
-                            ),
+                            trie_root: trie_root.to_vec().into(),
                             timestamp_sec: NonZeroU64::new(*timestamp).unwrap(),
                         };
                         AnyConsensusState::Rollup(state)
--- a/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
@@ -82,22 +82,24 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
     ) -> Result<Self::AnyConsensusState> {
         let store = self.borrow();
         #[cfg(feature = "witness")]
-        let state = (height.revision_number() == 1).then(|| {
-            store
-                .private
-                .local_consensus_state
-                .iter()
-                .find(|cs| cs.0 == height.revision_height())
-                .map(|(_slot, timestamp, trie_root)| {
-                    let state = cf_solana::ConsensusState {
-                        trie_root: ibc::CommitmentRoot::from_bytes(
-                            trie_root.as_slice(),
-                        ),
-                        timestamp_sec: NonZeroU64::new(*timestamp).unwrap(),
-                    };
-                    AnyConsensusState::Rollup(state)
-                })
-        });
+        let state = (height.revision_number() == 1)
+            .then(|| {
+                store
+                    .private
+                    .local_consensus_state
+                    .iter()
+                    .find(|cs| cs.0 == height.revision_height())
+                    .map(|(_slot, timestamp, trie_root)| {
+                        let state = cf_solana::ConsensusState {
+                            trie_root: ibc::CommitmentRoot::from_bytes(
+                                trie_root.as_slice(),
+                            ),
+                            timestamp_sec: NonZeroU64::new(*timestamp).unwrap(),
+                        };
+                        AnyConsensusState::Rollup(state)
+                    })
+            })
+            .flatten();
         #[cfg(not(feature = "witness"))]
         let state = (height.revision_number() == 1)
             .then(|| {
--- a/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
@@ -82,39 +82,39 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
     ) -> Result<Self::AnyConsensusState> {
         let store = self.borrow();
         #[cfg(feature = "witness")]
-        {
-            let (_slot, fetched_timestamp, fetched_trie_root) = store
-                .private
-                .local_consensus_state
-                .iter()
-                .find(|cs| cs.0 == height.revision_height())
-                .unwrap();
-            let rollup_consensus_state = cf_solana::ConsensusState {
-                trie_root: ibc::CommitmentRoot::from_bytes(
-                    fetched_trie_root.as_slice(),
-                ),
-                timestamp_sec: NonZeroU64::new(*fetched_timestamp).unwrap(),
-            };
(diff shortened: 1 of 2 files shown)


solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -79,27 +81,40 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
...
+            .ok_or(
+                ibc::ClientError::MissingLocalConsensusState {
+                    height: *height,
+                },
+            )?;
+            let guest_consensus_state = cf_guest::ConsensusState {
+                block_hash: state.0.as_array().to_vec().into(),
+                timestamp_ns: state.1,
+            };
+            return Ok(AnyConsensusState::Guest(guest_consensus_state));
```
mina86, line 116:
```suggestion
            Ok(AnyConsensusState::Guest(guest_consensus_state))
```

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -79,27 +81,40 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
...
-        } else {
-            None
+        #[cfg(feature = "witness")]
+        {
+            let (_slot, fetched_timestamp, fetched_trie_root) = store
+                .private
+                .local_consensus_state
+                .iter()
+                .find(|cs| cs.0 == height.revision_height())
+                .unwrap();
```
mina86, line 91: This should return an error rather than panicking, no? This can be handled with simple ok_or

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -79,27 +81,40 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
...
+                ibc::ClientError::MissingLocalConsensusState {
+                    height: *height,
+                },
+            )?;
+            let guest_consensus_state = cf_guest::ConsensusState {
+                block_hash: state.0.as_array().to_vec().into(),
+                timestamp_ns: state.1,
+            };
+            return Ok(AnyConsensusState::Guest(guest_consensus_state));
         }
```
mina86, line 117: Generation of the error could be move outside of the conditional compilation, though I’m not sure if this is better or worse. Up to you:

```suggestion
        #[cfg(feature = "witness")]
        let state = store
            .private
            .local_consensus_state
            .iter()
            .find(|cs| cs.0 == height.revision_height())
            .map(|(_slot, timestamp, trie_root)| {
                let state = cf_solana::ConsensusState {
                    trie_root: ibc::CommitmentRoot::from_bytes(
                        trie_root.as_slice(),
                    ),
                    timestamp_sec: NonZeroU64::new(*timestamp).unwrap(),
                };
                AnyConsensusState::Rollup(state)
            });
        #[cfg(not(feature = "witness"))]
        let state = (height.revision_number() == 1)
            .then(|| {
                let height = height.revision_height().into();
                store.chain.consensus_state(height)
            })
            .transpose()?
            .map(|state| {
                let state = cf_guest::ConsensusState {
                    block_hash: state.0.as_array().to_vec().into(),
                    timestamp_ns: state.1,
                };
                AnyConsensusState::Guest(state)
            });
        state.ok_or(ibc::ClientError::MissingLocalConsensusState {
            height: *height,
        })
```

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
```diff
@@ -79,27 +81,40 @@ impl ibc::ValidationContext for IbcStorage<'_, '_> {
         height: &ibc::Height,
     ) -> Result<Self::AnyConsensusState> {
         let store = self.borrow();
-        let state = if height.revision_number() == 1 {
-            store.chain.consensus_state(height.revision_height().into())?
-        } else {
-            None
+        #[cfg(feature = "witness")]
+        {
+            let (_slot, fetched_timestamp, fetched_trie_root) = store
```
mina86, line 86: Should we check height.revision_number here as well?

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: this looks clean.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: yes i think we should.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
mina86: Also, CryptoHash has to_vec function so it might be possible to do:

    trie_root: trie_root.to_vec().into(),

which is more concise.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
mina86: It might be easier to define a helper method which returns
`Result<Option<Self::AnyConsensusState>>` and have this use that:

        fn host_consensus_state(
            &self,
            height: &ibc::Height,
        ) -> Result<Self::AnyConsensusState> {
            self.host_consensus_state_impl(height).and_then(|state| {
                state.ok_or(ibc::ClientError::MissingLocalConsensusState {
                    height: *height,
                })
            })
        }

    // ...

        #[cfg(feature = "witness")]
        fn host_consensus_state_impl(
            &self,
            height: &ibc::Height,
        ) -> Result<Option<Self::AnyConsensusState>> {
            if height.revision_number() != 0 {
                return Ok(None);
            }
            Ok(self.borrow()
                .private
                .local_consensus_state
                .iter()
                .find(|cs| cs.0 == height.revision_height())
                .map(|(_slot, timestamp, trie_root)| {
                    AnyConsensusState::Rollup(cf_solana::ConsensusState {
                        trie_root: trie_root.to_vec().into(),
                        timestamp_sec: NonZeroU64::new(*timestamp).unwrap(),
                    })
                }))
        }

        #[cfg(not(feature = "witness"))]
        fn host_consensus_state_impl(
            &self,
            height: &ibc::Height,
        ) -> Result<Option<Self::AnyConsensusState>> {
            if height.revision_number() != 1 {
                return Ok(None);
            }
            let height = height.revision_height().into();
            self.borrow().chain.consensus_state(height).map(|state| {
                let state = cf_guest::ConsensusState {
                    block_hash: state.0.to_vec().into(),
                    timestamp_ns: state.1,
                };
                Some(AnyConsensusState::Guest(state))
            })
        }

Again, up to you.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: yes it should, solved it by propagating the error outside as u said.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
mina86: Hmm, actually, since we check for revision_number == 1 in both cases, that could be moved outside of the _impl method as well. That would potentially eliminate use of Option::then.

solana/solana-ibc/programs/solana-ibc/src/validation_context.rs
dhruvja: got it

PR #404, opened 2024-11-06
hooks: fix hook structure by deserializing json instead of string

dhruvja:
Since PFM only support passing of the memo for the next transfer as json, strings can be passed. So we have to deserialize the memo with json and get the string from it. Also log the acknowledgement after the hooks is performed since the ack can change based on the result of hooks.

### DIFF
--- a/solana/solana-ibc/programs/solana-ibc/src/transfer/mod.rs
+++ b/solana/solana-ibc/programs/solana-ibc/src/transfer/mod.rs
@@ -151,7 +151,6 @@ impl ibc::Module for IbcStorage<'_, '_> {
             );
         let ack_status = str::from_utf8(ack.as_bytes())
             .expect("Invalid acknowledgement string");
-        msg!("ibc::Packet acknowledgement: {}", ack_status);
 
         let status =
             serde_json::from_str::<ibc::AcknowledgementStatus>(ack_status);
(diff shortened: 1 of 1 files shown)


solana/solana-ibc/programs/solana-ibc/src/transfer/mod.rs
```diff
@@ -464,63 +470,75 @@ fn call_bridge_escrow(
...
-fn parse_bridge_memo(memo: &str) -> Option<(&str, &str)> {
-    let (count, mut memo) = memo.split_once(',')?;
+/// Memo is a JSON object with a `memo` field of the form
+/// `{ memo: "N,account-0,account-1,...,account-N-1,intent-id,embedded-memo" }`.
+/// Embedded memo can contain commas.  Returns `intent-id` and `embedded-memo`
+/// or `None` if the memo does not conform to this format.  Note that no
+/// validation on accounts is performed.
+fn parse_bridge_memo(memo: &str) -> Option<(String, String)> {
+    let parsed = serde_json::from_str::<serde_json::Value>(memo).ok()?;
+    let memo_str = parsed.get("memo")?.as_str()?.to_string();
```
mina86, line 480: `.to_string()` shouldn’t be necessary.

mina86 (approved):

solana/solana-ibc/programs/solana-ibc/src/transfer/mod.rs
dhruvja: oh yes, ill remove it.

drsadeghian (approved):

